PR #11, opened 2015-04-30
Remove activerecord-deprecated_finders

h-lame:
Mostly just updating ``find_by_blah(xxx)`` to ``find_by(blah: xxx)``.  We also rename ``find_by_code`` on RejectionReason to ``for_code`` because it's not an AR object and I don't want it to look like one (making people thing they can do ``find_by(other: attributes)``).


otlaitil:
LGTM

pixeltrix:
:shipit:

PR #14, opened 2015-05-01
Create data model for Sponsors feature

davidbasalla:
Added email uniqueness validation per peitition, allowing one email address
per petition only. May introduce more complex email address validation
later (similar to Signature validation)

### DIFF
--- a/lib/email_downcaser.rb
+++ b/lib/email_downcaser.rb
@@ -1,0 +1,8 @@
+class EmailDowncaser
+  def self.dump(value)
+    value.downcase
+  end
+  def self.load(value)
+    value
+  end
+end
--- a/app/models/petition.rb
+++ b/app/models/petition.rb
@@ -52,6 +52,7 @@ class Petition < ActiveRecord::Base
   accepts_nested_attributes_for :creator_signature
   has_many :signatures
   has_many :department_assignments
+  has_many :sponsors
 
   # = Validations =
   validates_presence_of :response, :if => :email_signees, :message => "must be completed when email signees is checked"
--- a/app/models/signature.rb
+++ b/app/models/signature.rb
@@ -24,19 +24,11 @@ class Signature < ActiveRecord::Base
 
   before_create :set_perishable_token
 
-  class EmailDowncaser
-    def self.dump(value)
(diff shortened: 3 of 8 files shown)


.gitignore
```diff
@@ -19,3 +19,8 @@ coverage
 */.DS_Store
 */*/.DS_Store
 .idea
+*/#*
+*/.#*
+#*
+.#*
+*~
```
pixeltrix: Why are these being added to .gitignore?

davidbasalla: Those are for temporary Emacs files, will remove them now.

lib/email_downcaser.rb
```diff
@@ -0,0 +1,8 @@
+class EmailDowncaser
+  def self.dump(value)
+    value.downcase
+  end
+  def self.load(value)
+    value
+  end
+end
```
pixeltrix: Not sure why we're down casing email addresses - technically the part before the @ can be case sensitive.

.gitignore
pixeltrix: Those kind of ignores should go in your global gitignore file

lib/email_downcaser.rb
davidbasalla: Shall I change it to only downcase the part after the @ per email?
This method was being used in the Signature model, so I moved it out into a separate file so that I can reuse it for Sponsor model as well.

h-lame: The bit in-front of the domain can be case sensitive, but I suspect it's very unlikely that it ever is.

I think it's to check against me using MURRAY.STEELE @ my domain and murray.steele @ my domain so I'd leave it as-is.  It's also likely that when we were storing the email address rather than the encrypted version we would not have been doing things in mysql to mean our "you've already signed this petition" lookups would have been case-sensitive, so downcasing the whole email address keeps that behaviour.

h-lame:
This seems basically fine to me.  I might be inclined to extract the shared behaviour between signature and sponsor to a module that we include as it's just complex enough (particularly with having to define the attribute methods to stop ``attr_encrypted`` breaking) to want us to keep it in one place.

PR #15, opened 2015-05-01
Get rid of captcha properly

h-lame:
Instead of leaving it in but behind a feature flag that means it's not in the UI and always verifies whatever text you give it we actually remove it fully.


otlaitil:
LGTM :+1:

PR #16, opened 2015-05-01
Some tidy up

h-lame:
Removing things we don't need, shuffling things around to be in more expected places for recent rails app infrastructure.


pixeltrix:
This will conflict #17 because of the signatures_pre_encryption table changes - we'll need to change either this one or the other one depending on which gets merged first.

Also, can we now remove `lib` from the `autoload_paths` array?

h-lame:
Rebased to address db conflicts with #17 and remove lib from autoload_paths.  Will merge when green.

PR #18, opened 2015-05-05
Refactor sponsor and signature models

davidbasalla:
Extracted the encrypted_email and set_perishable_token functionality out of sponsor and signature models and put them into separate modules


app/models/concerns/set_perishable_token.rb
```diff
@@ -0,0 +1,14 @@
+module SetPerishableToken
```
otlaitil: Maybe just nitpicking; I'm not sure if ```SetPerishableToken``` is a good name for a module. Maybe just ```PerishableToken``` instead?

otlaitil:
Looks good! :+1:

app/models/signature.rb
```diff
@@ -18,22 +18,25 @@
 #
 
 class Signature < ActiveRecord::Base
+
+  include EmailEncrypted
+  include SetPerishableToken
```
pixeltrix: I agree with @otlaitil - the names could be better - `EmailEncrypter` and `PerishableTokenGenerator` maybe?

PR #20, opened 2015-05-06
Collect email addresses of petition sponsors

otlaitil:


app/models/petition.rb
```diff
@@ -105,6 +107,23 @@ class Petition < ActiveRecord::Base
     where("signature_count >= ?", SystemSetting.value_of_key(SystemSetting::GET_AN_MP_SIGNATURE_COUNT).to_i)
   }
 
+  def sponsor_emails
+    @sponsor_emails || []
+  end
+
+  def sponsor_emails=(emails)
```
h-lame: I'd consider making this method assume the emails is an array already and do the string->array massaging in the controller.

app/views/petitions/new.html.erb
```diff
@@ -75,6 +76,25 @@
     <%= f.fields_for :creator_signature do |csf| %>
       <%= render :partial => 'signatures/form', :locals => { :f => csf, :signature => @petition.creator_signature } %>
     <% end %>
+  </fieldset>
+
+  <fieldset>
+    <% increment(2) # Skip 2 tabindexes for next/back buttons -%>
+    <legend>Sponsor email adresses</legend>
+    <div class="strap_block">
+     <%= form_row :for => [@petition, :sponsor_emails] do %>
```
h-lame: Boring nitpick: prefer 2-space indent.

spec/models/petition_spec.rb
```diff
@@ -52,6 +52,30 @@
...
+        expect(FactoryGirl.build(:petition, sponsor_emails: sponsor_emails)).not_to be_valid
+      end
+
+      it 'is not valid with invalid sponsor emails' do
+        sponsor_emails = ['test1test', 'test2@test.com', 'test3@test.com', 'test4@test.com', 'test5@test.com']
+        expect(FactoryGirl.build(:petition, sponsor_emails: sponsor_emails)).not_to be_valid
+      end
+
+      it 'is not valid with duplicate sponsor emails' do
+        sponsor_emails = ['test1@test.com', 'test1@test.com', 'test3@test.com', 'test4@test.com', 'test5@test.com']
```
h-lame: If I add a 6th email that is unique to this list the petition will be valid.

spec/models/petition_spec.rb
```diff
@@ -52,6 +52,30 @@
...
     it { is_expected.to validate_presence_of(:creator_signature).with_message(/must be completed/) }
 
+    context "sponsor email validations" do
+
+      it 'is valid with 5 sponsor emails' do
+        sponsor_emails = ['test1@test.com', 'test2@test.com', 'test3@test.com', 'test4@test.com', 'test5@test.com']
+        expect(FactoryGirl.build(:petition, sponsor_emails: sponsor_emails)).to be_valid
+      end
+
+      it 'is not valid with less than 5 sponsor emails' do
```
h-lame: What about ``is not valid with more than 20 sponsor emails`` ?

otlaitil: It is valid, it will just take the fist 20 of the sponsor emails and discard the rest.

app/models/petition.rb
```diff
@@ -220,4 +239,15 @@ def signature_counts_by_postal_district
       end
     end
   end
+
+  def validate_number_of_sponsor_emails
+    if sponsor_emails.count < AppConfig.sponsor_count_min
```
h-lame: As per my comment on the specs (see below), consider adding a validation that we're not trying to add more than ``AppConfig.sponsor_count_max``.  As it stands when someone does try to add more than that we're going to silently truncate their list and they may not realise.  It's probably better to let them choose which of the extra emails to remove rather than us just picking the first 20 of them.

app/models/petition.rb
```diff
@@ -105,6 +107,23 @@ class Petition < ActiveRecord::Base
...
+    @sponsor_emails || []
+  end
+
+  def sponsor_emails=(emails)
+    if emails.respond_to?(:each)
+      @sponsor_emails = emails
+    else
+      @sponsor_emails = parse_emails(emails)
+    end
+    @sponsor_emails = @sponsor_emails.select{ |e| Authlogic::Regex.email =~ e }.uniq.first(AppConfig.sponsor_count_max)
```
h-lame: Could we re-use ``Sponsor`` here rather than doing the validation "inline" and hoping that this method and Sponsor's validations remain in sync?

spec/models/petition_spec.rb
h-lame: Because the validation that fails here is not about duplicates, but about the number of emails being less than 5.  We should decide what the real rules here are.  Perhaps this test should be ``removes duplicate sponsor emails from the list`` rather than ``is not valid with duplicate sponsor emails``.

otlaitil: I agree, the test descriptions are incorrect.

h-lame:
I think we're ~80% here with this.  I think we could leverage the ``Sponsor`` and ``has_many :sponsors`` association more than we are.

app/models/petition.rb
otlaitil: Yes, this would be a better way to do it indeed! I will try to refactor it.

h-lame:
Yeah, this feels better, nice work @otlaitil!

h-lame:
Looks like we need a rebase against master (perhaps wait for #23) and then we can merge once green.

otlaitil:
@h-lame yep, I will do that. Is it ok if I squash all the commits in this branch to a single one before merging?

PR #21, opened 2015-05-07
Break up petition form into stages

h-lame:
Petition creation was a single form that we broke up into stages using JS.  This turns it into multiple server page refreshes.  We still only use new/create actions, but manage the process via a new ``StagedPetitionCreator`` object that works out the state.  We pass all the details of the object around in hidden fields and only persist to the DB once all stages have been filled in and we do our final submit.  We track the user action too so we can let them go back or forward along the stages.

I'm not 100% convinced that this is the correct approach, as I think if I was doing this from nothing I might have separate ``ActiveModel`` objects for each stage and only use a ``Petititon`` object on the final submit (I haven't thought how we'd share the validations properly, but I'm sure we could do something).  However, as an evolution from what we had this seems to work and I haven't been able to get into any weird edgecases with going back and forward.

Very much open to discussion, and I'm not against starting again with a different approach if we don't like this.


alanth:
The validation error "You must accept the terms and conditions." before I even get a chance to tick the box

h-lame:
Boo!  I'll look into it, but it might be something that pushes me towards a different implementation (which is not to say we shouldn't).

alanth:
In fact, it's probably not worthwhile fixing that validation error error, as an upcoming story removes the checkbox

app/helpers/petition_helper.rb
```diff
@@ -0,0 +1,32 @@
+module PetitionHelper
+  def render_petition_form(petition, form)
+    [
+      render_hidden_details(petition, form),
+      render_ui(petition, form)
+    ].join.html_safe
```
pixeltrix: Could probably use [`safe_join`][1] here - though can't we just use `capture` and `concat`?

[1]: http://api.rubyonrails.org/classes/ActionView/Helpers/OutputSafetyHelper.html#method-i-safe_join

app/helpers/petition_helper.rb
```diff
@@ -0,0 +1,32 @@
...
+  def render_hidden_details(petition, form)
+    out = []
+    out << hidden_field_tag(:stage, petition.stage)
+    out << render('/petitions/create/petition_details_hidden', petition: petition, f: form) unless petition.stage == 'petition'
+    if petition.creator_signature.present?
+      out << render('/petitions/create/your_details_hidden', petition: petition, f: form) unless petition.stage == 'creator'
+      out << render('/petitions/create/submit_hidden', petition: petition, f: form) unless petition.stage == 'submit'
+    end
+    out.join.html_safe
+  end
```
pixeltrix: Same here as above

pixeltrix:
Did you look at using an gem solution like [Wicked][1] ?

[1]: https://github.com/schneems/wicked

app/helpers/petition_helper.rb
h-lame: Ah, yes. I tried `capture` but forgot that I needed `concat` too to make it work.  Would probably make the ``render_hidden_details`` method clearer too.  Thanks!

h-lame:
Didn't even look for a gem solution.  I'll cast an eye over it and see, just a bit wary of adding more dependencies - but perhaps it's easier / sheds new light on a better approach.

h-lame:
Have done a reworking.  I think it's simpler now and we have more objects.  I've extracted the validations for each stage into concerns that are included in the real models (petition & signature) but also included in active model wrappers that only validate what each "stage" of the petition creation needs.  We also have a set of stage objects in the ``StagedPetitionCreator`` that should make it easier to inject new stages (sponsor collection, etc) rather than trying to unpick what we had.

If this looks ok, I'll do a rebase to merge in the stuff we've changed on master already.

h-lame:
Rebased against master so it now includes the sponsor collection stuff.

h-lame:
Unless anyone objects, when this goes green I'm going to merge it.

PR #19, opened 2015-05-05
Sending emails for new petition sponsors

otlaitil:
The email needs copy from Pete before merging.

PR #22, opened 2015-05-07
Let sponsors sign petitions

h-lame:
As per: https://www.pivotaltracker.com/story/show/93425644


h-lame:
This is WIP - we probably want #19 to be merged so we can start our feature from the email rather than just grabbing the token out of the db.

otlaitil:
Looks good!

h-lame:
Rebased against all the merged stuff so we could start the journey properly and added a "sad path" feature.  If it's green and people like, this would be ready to merge.

PR #23, opened 2015-05-08
Fix email reminder spec

otlaitil:
Fixes an email reminder spec that was occasionally failing. The
spec was comparing arrays with '==' instead of asserting that
array included the wanted values.


otlaitil:
@h-lame good point!

PR #24, opened 2015-05-11
Exclude petition creator email from sponsors

otlaitil:
Add a validator for sponsor email to exclude the
petition creator email.


otlaitil:
This fixes the issue @alanth found in QA, https://www.pivotaltracker.com/story/show/93424754

PR #25, opened 2015-05-11
Add action field to petition

davidbasalla:
Added action attribute to petition model, with form field on front end. Also added rspec and cucumber tests for it.


davidbasalla:
Looks like the schema.rb got changed, currently investigating...

h-lame:
Apart from the already mentioned schema changes, this looks good to me.  Fix those and we can commit.  Separately we should work out what version of Postgres we're planning to deploy to, make sure we're all using that one, and regenerate a canonical new db/schema.rb from that version.

PR #29, opened 2015-05-12
Add petition with 5 sponsors to mod queue

davidbasalla:
So this is WIP, but would love to discuss this tomorrow with someone, especially with regards to the testing.

I've added the new "sponsored" state for petition model. The state is set through the sponsors controller when the required sponsor amount has been reached. For the frontend, the new "sponsored" state has trumped the "validated" state, meaning that when before admins would see 'validated' petitions, they now see 'sponsored' petitions. I've updated all the rspec and cucumber tests to reflect this change. However I'm concerned that there not enough tests left for the "validated" state. Also, I have only one rspec test for the controller that triggers the state change to "sponsored" so thinking it would be good to test this more thoroughly.


h-lame:
This PR is based ontop of #26.  Note that I have recently rebased #26 to a more recent master so we'll need to be careful about merging / rebasing this PR too.

PR #31, opened 2015-05-13
Remove petition filtering by department on admin todolist

davidbasalla:
Removed the scope for departments in the todolist controller and updated the rspec and cucumber tests.


h-lame:
Looks good.

PR #32, opened 2015-05-13
Refactor search functionality

otlaitil:
Refactors search functionality to get rid of custom sunspot search
code.

Department search is not 100% working (ie. search counts do not match
the results) and lack unit testing. That is intentional because we are
getting rid of the department functionality.

Preventing search with empty search term was also removed to
simplify the code. I saw no point preventing that, it is not an
expensive query and user can see all petitions anyway by clicking
a link in the site.


app/decorators/petition_search.rb
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  def search_term
+   @params[:q] || ''
+  end
+
+  def result_count_for_state(state)
+    @petition_search_counts ||= execute_result_counts_query
+    @facets ||= @petition_search_counts.facet(:state).rows
+    default = -> { OpenStruct.new(count: 0) }
```
h-lame: OpenStruct :(

app/views/search/_results.html.erb
```diff
@@ -1,12 +1,12 @@
 <div class="petition_index_view">
   <%= render 'search/result_tabs' %>
   <div class="petition_list">
-    <% if @petition_search.petitions.present? -%>
+    <% unless @petitions.to_a.empty? -%>
```
h-lame: EXTREME NITPICK: We've delegated ``empty?`` in ``PetitionSearch`` to the ``petition_results``, so we can probably drop the ``to_a`` here.

otlaitil: True!

spec/lib/search_spec.rb
```diff
@@ -1,176 +0,0 @@
...
-
-    context "configured to return less results" do
-      subject { Search.new(options.merge(:per_page => 5)) }
-      it "does indeed return less results" do
-        expect(sunspot).to receive(:paginate).with(:page => page, :per_page => 5)
-        subject.execute("apples")
-      end
-    end
-  end
-end
```
h-lame: :+1: getting rid of this spec makes me happy :)

spec/decorators/petition_search_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+    end
+    it 'returns result count for rejected state' do
+      expect(petitions.result_count_for_state('rejected')).to eq 1
+    end
+  end
+
+  describe 'search result ordering' do
+    it 'can sort results by petition title ascending' do
+      search_params = { q: 'petition', page: 1, state: 'closed', sort: 'title', order: 'asc' }
+      expect(PetitionSearch.new(search_params).to_a).to eq([closed_petition_first, closed_petition_second, closed_petition_third])
```
h-lame: Would it be worth making ``petition_results`` on ``PetitionSearch`` public?  It seems a bit odd to have ``.to_a`` on ``PetitionSearch`` be the trigger that runs the search.  I'm not massively for or against any chance here, just seems like an odd api is all.

app/decorators/petition_search.rb
```diff
@@ -0,0 +1,75 @@
...
+  def initialize(params)
+    @params = params
+  end
+
+  def state
+    State::SEARCHABLE_STATES.include?(@params[:state]) ? @params[:state] : 'open'
+  end
+
+  def search_term
+   @params[:q] || ''
```
h-lame: In the old implementation there was a bunch of search term sanitization that went on (start at ``Search#construct_search_string``).  Is it worth keeping that stuff?  At least for now as I assume that solr/sunspot won't do it for us automatically.  I know the implementation of any sanitisation will be different when we move to an alternate search backend, but we might still need to do something.

spec/decorators/petition_search_spec.rb
otlaitil: Yes, I had the same feeling about the api being a bit odd when writing these test cases :) making ```petition_results``` public would make it clearer indeed.

h-lame:
+:100: for the approach, it's much clearer now!  I just have a couple of minor points that require discussion.

spec/decorators/petition_search_spec.rb
h-lame: I think there'd be an argument for dropping the delegation stuff too and just using the results object directly in the views etc...

app/decorators/petition_search.rb
otlaitil: What's wrong with OpenStruct? :)

Do you find this better:

```@facets.find { |f| f.value.to_s == state }.try(:count) || 0```

h-lame: OpenStruct is great for prototyping, but shouldn't leak out into production code.  The implementation you have is nice (keen to avoid ``try`` if we can) - we just need a different thing to work as the default.  Could use an empty array, because that has ``count`` that returns ``0``, but it's a bit opaque.  I'd probably create an explicit ``NullFacet`` class that has the right API and returns ``0`` or empty string in all the right places.

otlaitil: Ok, let's go with ```NullFacet```. Thanks for the comments!

h-lame:
Looks good, alas we're burning through stories so fast it needs another rebase!

PR #33, opened 2015-05-14
No more departments for admin users

h-lame:
For: https://www.pivotaltracker.com/story/show/93906180

Admin users no longer have a relationship to departments.  This required a bit of tidy up where the app still assumed they did.  Tidying up the features was a bit brute force to make them pass.  The features will benefit from a fuller review of the steps and scenarios once we've completely removed departments from the app as I'm sure there could be more to cull.

### DIFF
--- a/db/migrate/20150513162113_remove_admin_users_departments_join_table.rb
+++ b/db/migrate/20150513162113_remove_admin_users_departments_join_table.rb
@@ -1,0 +1,5 @@
+class RemoveAdminUsersDepartmentsJoinTable < ActiveRecord::Migration
+  def change
+    drop_table 'admin_users_departments'
+  end
+end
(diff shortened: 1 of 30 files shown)


otlaitil:
:+1:

db/migrate/20150513162113_remove_admin_users_departments_join_table.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveAdminUsersDepartmentsJoinTable < ActiveRecord::Migration
+  def change
+    drop_table 'admin_users_departments'
```
pixeltrix: Not worth changing but there is a `drop_join_table :admin_users, :departments`

pixeltrix:
:shipit:

db/migrate/20150513162113_remove_admin_users_departments_join_table.rb
h-lame: Oh, I thought I knew all the secret migration helpers.  That's a new one to me.  Thanks!

PR #36, opened 2015-05-14
Remove address and town fields from front end and back end

davidbasalla:
I removed the address and town fields for the petition creator and signer, as well as all the related backend code and tests for the town and address fields.


h-lame:
:+1:

yahoopete:
+6 -59
:shipit:

PR #37, opened 2015-05-14
Dont ask for a department for a petition

h-lame:
For: https://www.pivotaltracker.com/story/show/93907142

I haven't removed the department_id from petition, just the UI to let you create it.  I think once all the other department removal stories are in we can do a final sweep to actually remove the associations and the ``Department`` model entirely.


otlaitil:
LGTM. :shipit:

PR #38, opened 2015-05-15
Fix sponsors validation during petition creation

h-lame:
To fix issue raised against https://www.pivotaltracker.com/story/show/93424754 (although actually caused by https://www.pivotaltracker.com/story/show/93703696)

### DIFF
--- a/app/models/staged/sponsors.rb
+++ b/app/models/staged/sponsors.rb
@@ -15,10 +15,8 @@ module Staged
     private
 
     def sponsors_valid?
-      # Ask them all to validate...
-      sponsors.each { |sponsor| sponsor.valid? }
-      # ...return if any have an error
-      sponsors.any? { |sponsor| sponsor.errors.any? }
+      # Make sure we ask all sponsors to validate before returning
+      sponsors.map(&:valid?).all?
     end
   end
 end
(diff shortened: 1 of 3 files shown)


otlaitil:
:+1:

app/models/staged/sponsors.rb
```diff
@@ -19,8 +17,8 @@ def valid?
     def sponsors_valid?
       # Ask them all to validate...
       sponsors.each { |sponsor| sponsor.valid? }
-      # ...return if any have an error
-      sponsors.any? { |sponsor| sponsor.errors.any? }
+      # ...return true if there are no errors at all
+      sponsors.all? { |sponsor| sponsor.errors.none? }
```
pixeltrix: Can't this just be `sponsors.all?(&:valid?)`

pixeltrix:
Other than the comment :+1:

app/models/staged/sponsors.rb
h-lame: I thought so, but it would shortcut and stop at the first ``false``, whereas I want to explicitly call ``valid?`` on each sponsor, so that if we get all the errors for displaying in the UI.

otlaitil: ```sponsors.map(&:valid?).all?``` ? The original version may be clearer though.

h-lame: Oh... that is neater!  Good call.  I think the version we have now is too verbose, I'll change it.

PR #39, opened 2015-05-15
Remove notify_by_email field and set default to true

davidbasalla:
Set the default value of notify_by_email to be 'true'. Removed the html for querying the user.

### DIFF
--- a/spec/controllers/signatures_controller_spec.rb
+++ b/spec/controllers/signatures_controller_spec.rb
@@ -160,6 +160,11 @@ describe SignaturesController do
           expect(assigns(:signature).petition).to eq(petition)
         end
 
+        it "has not changed the default TRUE value for notify_by_email" do
+          do_post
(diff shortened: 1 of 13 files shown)


spec/controllers/signatures_controller_spec.rb
```diff
@@ -160,6 +160,11 @@ def do_post(options = {})
           expect(assigns(:signature).petition).to eq(petition)
         end
 
+        it "has not changed the default TRUE value for notify_by_email" do
```
davidbasalla: Was wondering if this is an appropriate test to add at this stage? I want to check that somehow the default "notify_by_email" attribute has not changed somehow through the signing process...

h-lame: I think it's reasonable.  Here using ``eq true`` feels ok because we actually do want the value to be literally ``true`` or literally ``false``.

h-lame:
Looks good.

PR #40, opened 2015-05-15
Petitions have fixed durations

h-lame:
As per: https://www.pivotaltracker.com/story/show/94483442

### DIFF
--- a/app/models/petition.rb
+++ b/app/models/petition.rb
@@ -155,6 +155,13 @@ class Petition < ActiveRecord::Base
     counts_by_state
   end
 
+  def publish!
+    self.state = Petition::OPEN_STATE
+    self.open_at = Time.zone.now
+    self.closed_at = AppConfig.petition_duration.months.from_now.end_of_day
+    save!
+  end
+
   def reassign!(new_department)
     self.department = new_department
     save!
--- a/app/models/petition.rb
+++ b/app/models/petition.rb
@@ -20,7 +20,6 @@
 #  internal_response       :text
 #  rejection_code          :string(50)
 #  notified_by_email       :boolean(1)      default(FALSE)
-#  duration                :string(2)       default("12")
 #  email_requested_at      :datetime
 #  get_an_mp_email_sent_at :datetime
 #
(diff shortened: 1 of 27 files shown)


app/models/petition.rb
```diff
@@ -156,6 +155,13 @@ def self.counts_by_state
     counts_by_state
   end
 
+  def publish!
+    self.state = Petition::OPEN_STATE
+    self.open_at = Time.zone.now
```
pixeltrix: Why the instance on `Time.zone.now` vs. `Time.current` ?

app/models/petition.rb
```diff
@@ -156,6 +155,13 @@ def self.counts_by_state
...
+    self.closed_at = AppConfig.petition_duration.months.from_now.end_of_day
```
pixeltrix: Hmm, I wonder whether fractional timestamps might bite us here - worth checking

pixeltrix:
:shipit:

app/models/petition.rb
h-lame: Oh, good point.  I did have to handle it in the tests because the in-memory time was ``23:59:59.99999999`` but the db roundtrip gave us ``23:59:59.99990000`` (for some estimate of decimal places).  Any ideas?

h-lame: Could change it to ``(AppConfig.petition_duration.months.from_now + 1.day).midnight`` which would be 1 second after ``AppConfig.petition_duration.months.from_now.end_of_day`` with the advantage of 0-ing all the hours, minutes, seconds, and fractions.

pixeltrix: How is the comparison done? If we're doing `<` then we can do what you suggested.

app/models/petition.rb
h-lame: I blindly copied it from the controller impl. ;)  But you're right, I'll change it and do a sweep through the rest of the app too.

app/models/petition.rb
h-lame: It is ``<``, but on relfection I think I'll leave it as-is.  If we make it the ``+ 1.day).midnight`` version then we'll have to do weird things when displaying the closing date.  It might be correct that midnight is the start of the day, not the end of the day, but I think if you said to someone that the petition closed at midnight on a given day, they'd think the end, not the start.  So it's easier to have the ``.end_of_day`` version stored, than the ``.midnight`` version in terms of display logic.

PR #41, opened 2015-05-15
Change UK resident radio button into checkbox

davidbasalla:
Changed the radio button into a checkbox and adjusted the tests. I couldn't actually find anywhere in the database that saves this 'citizenship' attribute so have not done anything on the backend.

### DIFF
--- a/app/views/signatures/_form.html.erb
+++ b/app/views/signatures/_form.html.erb
@@ -24,19 +24,11 @@
   <% end %>
 <% end %>
 
-<%= form_row :class => 'is_uk_citizen_row', :for => [signature, :uk_citizenship] do %>
-  <p class="label">British citizen or UK resident?</p>
-  <ul class="radio_button_list">
-    <li>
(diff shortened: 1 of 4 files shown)


h-lame:
It'll be saved semi-automatically because it's in the params we get from the controller and there is a field in the db for it.  We agreed that we would remove the field from the db in this story so we'll need a migration to remove it and then add an ``attr_accessor :uk_citizenship`` to put it back as a "virtual attribute".  Much like we used to have with ``town`` and ``address`` on ``Signature``, before we dropped them.

davidbasalla:
Hmmm, sorry still can't find the darn thing. I'm looking in db/schema.rb for a 'citizen' boolean field but not finding it? Should I be looking somewhere else?

h-lame:
You're right, I'm wrong.  It already is a virtual attribute, I was thinking back to when @alanth said all the values in the db were ``1``, but then we realised we had crossed-wires and were talking about different stories and different boolean flags.  Ignore my comment.  Erase your browser history.  Nothing to see here.

PR #42, opened 2015-05-18
Remove emailing creator when reaching mp notification threshold

davidbasalla:
Have removed the cron job and petition method that was getting called to email creator once reaching the find_mp_threshold, as a well as the related testing code. I have NOT removed the scope "eligible_for_get_an_mp_email" as I understand this might still be used in some way later (with the 10k signature milestone).


davidbasalla:
Cheers, I've pushed a new commit to address those comments!

h-lame:
Is good.

otlaitil:
:+1:

PR #45, opened 2015-05-19
Changing regex validation on email so it accepts top level domain addresses

mybits:
PT: https://www.pivotaltracker.com/story/show/94486998

### DIFF
--- a/app/controllers/petitions_controller.rb
+++ b/app/controllers/petitions_controller.rb
@@ -47,7 +47,7 @@ class PetitionsController < ApplicationController
 
   def resend_confirmation_email
     @petition = Petition.visible.find(params[:id])
-    SignatureConfirmer.new(@petition, params[:confirmation_email], PetitionMailer, Authlogic::Regex.email).confirm!
+    SignatureConfirmer.new(@petition, params[:confirmation_email], PetitionMailer, EMAIL_REGEX).confirm!
   end
 
   protected
@@ -112,4 +112,3 @@ class PetitionsController < ApplicationController
     emails.strip.split(/\r?\n/).map { |e| e.strip }
   end
 end
-
--- a/app/models/concerns/email_encrypter.rb
+++ b/app/models/concerns/email_encrypter.rb
@@ -14,13 +14,12 @@ module EmailEncrypter
     validates_presence_of :email, message: "%{attribute} must be completed"
 
     validates_format_of :email,
-                        with: Authlogic::Regex.email,
+                        with: EMAIL_REGEX,
                         unless: 'email.blank?',
                         message: "Email '%{value}' not recognised."
-    
+
     # = Finders =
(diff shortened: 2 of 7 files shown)


h-lame:
There's a part of me that thinks a couple of specs around emails we do and don't recognize might be useful.  I know we don't really have them already, but that in part would be because we were using the Authlogic regex and want to assume it does it's job.  Now we've got our own regex I think it might be a good idea to at least set out our intention even if no spec will cover all the possibilities that might arise.

I'm willing to be swayed from this if no-one else feels strongly though.

davidbasalla:
Hooray for more tests!

mybits:
I agree @h-lame, indeed using Authlogic allowed to skip more tests around email validation. I'll be happy to add some.

mybits:
Thanks for the comments! I've updated PR here: https://github.com/alphagov/e-petitions/commit/7996c39ebdccfd72978f1bcf15b3bbbbd2fcfd16

otlaitil:
@mybits I would consider making the email regex tests to be independent, to test just the regex pattern and nothing else.

otlaitil:
I mean, maybe something like here: https://github.com/balexand/email_validator/blob/master/spec/email_validator_spec.rb , what do you think?

mybits:
@otlaitil I like the example you posted, but do we actually need such a huge validator?

h-lame:
I'd be interested in a spec just for that regex, throwing a bunch of emails at it to see what it does and doesn't match.  Not super interested in repeating those specs across all the objects that should be using it for the validator.

Something like:

```ruby
describe "Email validation" do
  subject { EMAIL_REGEX }
  it "matches simple email addresses" do
    expect(subject).to match('laura@example.com')
  end
  it "matches subdomain email addresses" do
    expect(subject).to match('laura@subdomain.example.com')
  end
  it "doesn't match email addresses without a domain and tld" do
    expect(subject).not_to match('laura@example')
  end
  # lots more examples.... (the ones from your feedback spec would be fine)
end
```

Note that this if off the top of my head, depending on the available rspec matchers we might have to flip to ``expect('laura@example').not_to match(subject)``.

otlaitil:
@mybits I think we could have only the regex tests, ie:

```
[
  "uncommon-tld@sld.museum",
  "uncommon-tld@sld.travel",
  "uncommon-tld@sld.mobi",
  ...
].each do |email|
  it "#{email.inspect} should match the regexp" do
    expect(email =~ EMAIL_REGEXP).to be_truthy
  end
end
```

mybits:
Thanks for all comments! PR updated, I've created separate tests just for email validation here: https://github.com/alphagov/e-petitions/commit/13d2cf555b2316a4cd51bf324233743dc8da87c3

h-lame:
Nice.  Assuming :green_heart: I'm happy to see this merged.

otlaitil:
Looks good! :+1:  :shipit:

PR #47, opened 2015-05-19
Remove admin role

otlaitil:
Also removes an email notification / reminder that was sent for admin users.
The functionality was no longer needed

Removes a separate view for internal response in admin
view. This was for admin users. All the changes to petitions
are now done from edit response -view.

Renames threshold role to moderator


spec/factories.rb
```diff
@@ -15,8 +15,8 @@
     role "sysadmin"
   end
 
-  factory :threshold_user, :parent => :admin_user do
-    role "threshold"
+  factory :moderator_user, :parent => :admin_user do
```
otlaitil: This moderator factory is probably redundant. Should I declare admin factory with

```factory :admin, aliases: [:moderator] do```

and get rid of this?

h-lame: Maybe just take the role out of the admin_user factory, and make sure we always use sysadmin_user, or moderator_user ?

otlaitil: That's better. Thanks.

otlaitil:
PR comments addressed.

h-lame:
We could merge this now I think.

davidbasalla:
:+1:

PR #48, opened 2015-05-20
Removal of the LinkedIn sharing icon

yahoopete:
This was never used and was removed, not sure why it has crept back, but let's bin it.  I have left the image in the icon folder, justincase it gets super useful again in the future.


mybits:
LGTM

h-lame:
Delete the image too!  It's not like we can't get it back.

PR #49, opened 2015-05-20
Remove FAQs from the site

yahoopete:
This removes FAQ page and reference to the FAQ page from the service.  Also removes the styles, tests etc.

Delivers on https://www.pivotaltracker.com/story/show/95038030


mybits:
I put one comment on redundant css, otherwise looks good to me

yahoopete:
Cool, I've removed that block.

PR #50, opened 2015-05-20
Resend email to unvalidated signature

davidbasalla:
Have edited the create method in the signature controller to check if any matching signatures (with the form details) already exist and if yes, resend the email.

Otto helped me massively with the refactor but would still love to get some feedback. Thanks!


h-lame:
Note to @davidbasalla I've rebased this against master so be careful when you pull/push.  I'm keen to get it merged today, but I do think we need to add something to those cucumber scenarios around the signature count.  Let's discuss after standup.

h-lame:
Oops!  I merged this by mistake.  I've reset master so it's as if it never happened, but we'll have to reopen this PR as a new one.  Sorry!

h-lame:
Reopened as #52.

PR #51, opened 2015-05-20
Fix bug of displaying raw html code as text for repeating signature v…

davidbasalla:
…erification

moved the .html_safe into the view, added a cucumber test for it.

### DIFF
--- a/app/controllers/signatures_controller.rb
+++ b/app/controllers/signatures_controller.rb
@@ -39,7 +39,7 @@ class SignaturesController < ApplicationController
     else
       # We've found the signature, but it's already been verified.
       if @signature.state == Signature::VALIDATED_STATE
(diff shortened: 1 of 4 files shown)


h-lame:
This is good.  Spoke with @pixeltrix about an alternative approach where we put a key, not a string into the flash.  In the view instead of just rendering out the contents of the flash we do a lookup into the i18n locale files with that key.  We can name our locale keys with ``_html`` to get html_safe! behaviour for free (see: http://guides.rubyonrails.org/i18n.html#using-safe-html-translations).

That said.  It's something we'd want to do as a sweep through the whole application to make sure this is the approach we use for *all* our flash messages so it's not worth doing just here.  I've added this as a tech-debt story in pivotal as https://www.pivotaltracker.com/story/show/95133802.

PR #54, opened 2015-05-21
Update email regex logic in javascript files that handling feedback and signature forms

mybits:
This is to fix javascript forms that were still having old regex logic for email validation.


alanth:
LGTM

h-lame:
Worth pointing out that I've dropped the JS from the signature form in my branch for https://www.pivotaltracker.com/story/show/94485346 (incoming real soon now) so maybe we don't need to pay too close attention to this right now.

h-lame:
That said, this is the correct fix so we can merge it.

mybits:
Thanks for the comments! Merging!

PR #44, opened 2015-05-19
Replay email address instead of confirming during create

h-lame:
For: https://www.pivotaltracker.com/story/show/94485374

1. Remove email_confirmation from creating and sigining a petition
2. Add a new step to the create process to replay the email address as the last step before submitting and allow the signer to make changes

Note: replaying the email while signing a petition will be dealt with in a future PR.

PR #55, opened 2015-05-21
Replay email address during signing

h-lame:
For: https://www.pivotaltracker.com/story/show/94485346

As in #44 we want to replay the email address during petition signing.  This involves adding a staged object creation style to the signatures controller which is similar to the staged creation of petitions.  We take the opportunity to try and refactor / shuffle things around to avoid duplication.

One thing to note: if you are trying to sign again and the previous signature is unvalidated we skip past the replay-email step and go straight to the thanks page.  Hopefully this is ok.


mybits:
LGTM :maple_leaf:  great refactoring!

davidbasalla:
Looks good

app/models/staged/petition_creator.rb
```diff
@@ -0,0 +1,62 @@
...
+
+      private
+
+      def stage_manager
+        @_stage_manager ||= Staged::StageManager.new(stages, previous_stage, move, petition)
+      end
+
+      def sanitize!
+        if petition.creator_signature
+          petition.creator_signature.email.strip! unless petition.creator_signature.email.blank?
```
otlaitil: Just thinking, I know that this PR is merged, but for curiosity (and to make a note that I went through this PR code :) could we use ```petition.creator_signature.email.to_s.strip!``` and ```petition.title.to_s.strip!``` to get rid of these two ```unless``` clauses. Or is it a bad practise?

app/models/staged/petition_creator/has_creator_signature.rb
```diff
@@ -0,0 +1,30 @@
+module Staged
+  module PetitionCreator
+    module HasCreatorSignature
+      extend ActiveSupport::Concern
+
+      included do
+        def valid?
+          super && creator_signature_valid?
```
otlaitil: This uses short circuiting and does not validate creator's signature if ```super``` returns ```false```, we potentially lose the display of some error msgs; should it use eager evaluation instead?

h-lame: Argh!  So it does.  Good spot!  

We have the same problem in ``Staged::PetitionCreator::Sponsors#valid?`` too, which is weird because I went to great pains to avoid the same short-circuiting problem in ``Staged::PetitionCreator::Sponsors#sponsors_valid?``!

h-lame: Addressed in #59.

app/models/staged/petition_creator.rb
h-lame: Had a think about this and it would work because:

1. calling ``to_s`` on a string returns that string, not a copy (this might be the kind of thing that changes between ruby stdlib versions)
2. calling ``to_s`` on ``nil`` to "cast" the value to a string to safely call ``strip!`` on it.

It's a neat trick, but I don't think it's as intention revealing as being explicit about what we're doing.  There might be a debate about performance though; perhaps always calling ``.to_s.strip!`` is faster (no conditional check) or maybe it's slower (always doing it even when we don't need to).  Either way it'd be a micro-optimisation.

PR #56, opened 2015-05-22
Typo in sponsor signing petition agreement

mybits:
Just notice that typo, so here's quick fix


alanth:
LGTM, will merge when Travis has done it's thing

PR #57, opened 2015-05-22
Removing the crown copyright page - replace with link

yahoopete:
https://www.pivotaltracker.com/story/show/95039778

We don't need the page, but we do still need an external link, so I've replaced one with the other (not opening in new window deliberately).


mybits:
looks good to me

PR #59, opened 2015-05-26
Avoid short-circuiting ``valid?`` in Staged models

h-lame:
In the classes where we have overridden ``valid?`` to also validate a child object (the ``HasCreatorSignature`` mixin and ``Sponsors`` model) we were doing:

    def valid?
      super && some_other_method_that_calls_valid_on_a_child_object
    end

The problem here is that if ``super`` returned false we short-circuit the expression and don't run the other method and so don't validate the child object.  Instead we use ``validate`` to call the other method as part of the standard validation process.

In practice it wasn't a problem as I neither of the affected classes had any validations on the "parent" so ``super`` would never fail.  However it's not a good pattern to leave lying around.

### DIFF
--- a/app/models/staged/petition_creator/has_creator_signature.rb
+++ b/app/models/staged/petition_creator/has_creator_signature.rb
@@ -4,13 +4,13 @@ module Staged
       extend ActiveSupport::Concern
 
       included do
-        def valid?
-          super && creator_signature_valid?
-        end
+        validate :creator_signature_valid?
 
         def creator_signature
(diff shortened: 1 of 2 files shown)


h-lame:
Thanks to @otlaitil for spotting this on #55.

otlaitil:
:+1: !

davidbasalla:
Cool, good to know!

mybits:
:+1:

PR #60, opened 2015-05-26
Adding a cookies page

yahoopete:
As per: https://www.pivotaltracker.com/story/show/95040606

This PR adds the new cookie page with *probably* correct content, then adds it into all the usual places routes and adds a couple of tests.

### DIFF
--- a/spec/controllers/static_pages_controller_spec.rb
+++ b/spec/controllers/static_pages_controller_spec.rb
@@ -28,4 +28,10 @@ describe StaticPagesController do
       expect(privacy_policy_path).to eq "/privacy-policy"
     end
   end
+  describe "cookies" do
+    it "should respond to /cookies" do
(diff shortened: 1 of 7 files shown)


spec/controllers/static_pages_controller_spec.rb
```diff
@@ -28,4 +28,10 @@
       expect(privacy_policy_path).to eq "/privacy-policy"
     end
   end
+  describe "cookies" do
+    it "should respond to /cookies" do
+      expect({:get => "/cookies"}).to route_to({:controller => "static_pages", :action => "cookies"})
+      expect(cookies_path).to eq "/cookies"
+    end
+  end
```
pixeltrix: This is a routing spec and should live in `spec/routing` but since you're just adding to the existing ones, I'll let it slide :wink:

alanth:
Follows the pattern of existing code, meging

PR #62, opened 2015-05-26
Pre-creation searching petitions results limited to 3 ones

mybits:
So I've updated this PR as we're actually removing the petition state tabs for that particular search (thanks @otlaitil for look-out) and generally provide one view results. I've tried to follow what we're currently having in a design, meaning removing table, adding petition description, but not being very exact with styling as we're saying goodbye to the current css

PT: https://www.pivotaltracker.com/story/show/93884116

### DIFF
--- a/app/decorators/petition_search.rb
+++ b/app/decorators/petition_search.rb
@@ -22,6 +22,10 @@ class PetitionSearch
     @results ||= petitions.results
   end
 
+  def limited_results_for_create_petition_search
+    execute_search_for_create_petition_query.results
+  end
+
   def result_count_for_state(state)
     @facets ||= petition_result_counts.facet(:state).rows
     default = -> { NullFacet.new }
@@ -45,7 +49,6 @@ class PetitionSearch
   def execute_search_query
     Petition.search do |query|
       query.fulltext search_term_sanitised
-      query.facet :state
       query.paginate page: @params[:page], per_page: 20
       case state
       when State::CLOSED_STATE
@@ -57,6 +60,15 @@ class PetitionSearch
         query.with(:state).equal_to("open")
         query.with(:closed_at).greater_than(Time.current.utc)
       end
+      query.order_by *SearchOrder.sort_order(@params, [:score, :desc])
+    end
+  end
+
+  def execute_search_for_create_petition_query
+    Petition.search do |query|
+      query.fulltext search_term_sanitised
+      query.paginate page: 1, per_page: 3
+      query.without(:state, 'hidden')
       query.order_by *SearchOrder.sort_order(@params, [:score, :desc])
     end
   end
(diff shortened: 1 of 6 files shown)


app/decorators/petition_search.rb
```diff
@@ -22,6 +22,10 @@ def results
     @results ||= petitions.results
   end
 
+  def limited_results
+    @results ||= petitions.results[0..2]
```
otlaitil: how about ```petitions.results.take(3)```?

otlaitil:
Should this story combine the results from all three visible states, get rid of the three tabs on the search results page, and show all the results in a single tab? I think petition results at the moment contain only results matching a single state?

app/decorators/petition_search.rb
mybits: Thanks! That's definitely cleaner:), have already changed it.

mybits:
Thanks @otlaitil - I've followed ACs of the story and kept current functionality (except from a different number of results of course:)) but have already asked @yahoopete about it, so soon we should know

h-lame:
A problem we have here is that ``PetitionSearch`` only allows searching by a single state at the moment, so we're going to have to change it to allow specifying a bunch of states because we want to include open, closed, and rejected petitions as one list here.

mybits:
Thanks for your comments @h-lame and @otlaitil - I've updated the PR and decided to give it a go with a  new search method for pre-create petition search results, specs have been updated with new examples as well.

app/decorators/petition_search.rb
```diff
@@ -61,6 +65,15 @@ def execute_search_query
     end
   end
 
+  def execute_search_for_create_petition_query
+    Petition.search do |query|
```
otlaitil: Could we possibly limit the number of results here to make the Solr return only the limited number of results (could be more efficient)?

```query.limit 3``` could work.

Maybe even fetch the actual limit from ```AppConfig``` so it would be easily changeable? :)

Also, I am not sure if we need to use ```query.facet :state``` because we don't need to group the results by the state.

h-lame: I agree with @otlaitil; as we have a new search method we can do the limit in there rather than using ``.take(3)`` on the results.  Ditto the facet thing, we almost certainly don't need it.

mybits:
Yes absolutely!
The only problem is that `limit` method works only for groups: https://github.com/sunspot/sunspot#grouping

So instead of it I can only use sunspot's `paginate` method to limit the search results:
`query.paginate page: 1, per_page: 3`
(it works fine)
What do you think?

otlaitil:
LGTM! :shipit:

mybits:
Thanks @otlaitil and @h-lame for all your comments! PR is updated now if you could have a final look

h-lame:
@mybits, can you rebase against master and we'll merge it when :green_heart:

mybits:
:recycle: done :)

PR #63, opened 2015-05-26
Add constituency info to petition summary

davidbasalla:
Adding the constituency info to the petition replay page during petition creation. Created an API wrapper for querying the data.parliament API, using faraday gem. Also added webmock and faraday gems for testing the API.


davidbasalla:
Going to look at using some other gems instead of HTTParty and VCR today

features/charlie_creates_a_petition.feature
```diff
@@ -46,9 +47,37 @@ Scenario: Charlie creates our petition
   Then a petition should exist with title: "The wombats of wimbledon rock.", state: "validated"
   And there should be a "validated" signature with email "womboid@wimbledon.com" and name "Womboid Wibbledon"
 
+
+Scenario: Charlie creates a petition with valid postcode N1 1TY
+  Given I start a new petition
+  And ConstituencyAPI call N11TY has been stubbed
```
otlaitil: We should probably handle the stubbing some other way. This reveals the implementation details which integration tests should not be concerned of. And the test cases should be understandable for non technical people too (this will probably make no sense for a non-programmer).

davidbasalla: Agreed! Will get on it

app/models/signature.rb
```diff
@@ -65,4 +65,11 @@ def validated?
   def postal_district
     postcode.upcase[0..-4].match(/[A-Z]{1,2}[0-9]{1,2}[A-Z]?/).to_s
   end
+
+  def constituency
+    @constituency ||= ConstituencyApi::Client.constituencies(self.postcode).first
+  rescue ConstituencyApi::ConstituencyApiError
```
otlaitil: In here we could log the error message.

app/lib/constituency_api.rb
```diff
@@ -0,0 +1,51 @@
...
+      constituencies = response["Constituencies"]["Constituency"]
+      Array.wrap(constituencies).map { |c| Constituency.new(c["Name"]) }
+    end
+    
+    def self.call_api(postcode)
+      response = Faraday.new(URL).get("#{postcode_param(postcode)}/") do |req|
+        req.options[:timeout] = TIMEOUT
+        req.options[:open_timeout] = TIMEOUT
+      end
+      raise ConstituencyApiError.new("Unexpected response") unless response.status == 200
```
otlaitil: We could include some more information on this error message. When the API is changed (or there is a special case we haven't taken into account), our wrapper breaks and we would then have meaningful log messages to help us fix the issue. 

```ConstituencyApiError.new("Unexpected response from API: status #{response.status} body #{response.body}")```? And maybe even include the full request url in the message.

File: app/models/signature.rb (hunk 2 above)
Comment (davidbasalla) [inline app/models/signature.rb]: using this kind of thing for example?
@logger = AuditLogger.new(Rails.root.join('log', logfilename), 'email reminders')

Comment (otlaitil) [inline app/models/signature.rb]: I think it's something like
```
def constituency
  @constituency ||= ConstituencyApi::Client.constituencies(self.postcode).first
rescue ConstituencyApi::ConstituencyApiError => e
  Rails.logger.error('"Failed to fetch constituency", e)
end
```
@h-lame is that right?

File: app/lib/constituency_api.rb [hunk 4]
```diff
@@ -0,0 +1,51 @@
...
+    attr_accessor :name
+    
+    def initialize(name)
+      self.name = name
+    end
+    
+    def ==(another_constituency)
+      self.name == another_constituency.name
+    end
+  end
```
h-lame: It is excellent to have an actual data object, so +:100: for that!  As another bit of ruby knowledge we could write this as:

```ruby
class Constituency < Struct.new(:name)
end
```

And we'd get the initializer, accessor and equality for free.  It's worth exploring structs to see the pros-and-cons of using them.  Here's a starter blog post http://blog.steveklabnik.com/posts/2012-09-01-random-ruby-tricks--struct-new

features/charlie_creates_a_petition.feature
```diff
@@ -46,6 +46,29 @@ Scenario: Charlie creates our petition
...
+
+Scenario: Charlie creates a petition with valid postcode N1 1TY
+  Given I start a new petition
+  And I fill in the petition details
+  And I press "Next"
+  And I fill in my details with postcode "N1 1TY"
+  And I press "Next"
+  And I fill in sponsor emails
+  And I press "Next"
+  Then I should see my constituency "Islington South and Finsbury"
```
h-lame: Could probably roll this entire scenario into the existing "happy path" scenario.  Maybe.  Anyone else got an opinion?

features/step_definitions/petition_steps.rb
```diff
@@ -159,3 +159,26 @@
...
+  )
+end
+
+When /^I fill in the petition details/ do
+  steps %Q(
+    When I fill in "Title" with "The wombats of wimbledon rock."
+    And I fill in "Action" with "Give half of Wimbledon rock to wombats!"
+    And I fill in "Description" with "The racial tensions between the wombles and the wombats are heating up. Racial attacks are a regular occurrence and the death count is already in 5 figures. The only resolution to this crisis is to give half of Wimbledon common to the Wombats and to recognise them as their own independent state."
+  )
+end
```
h-lame: These two "meta" steps are neat, could we use them in other scenarios too?

spec/models/signature_spec.rb
```diff
@@ -18,6 +18,7 @@
 #
 
 require 'rails_helper'
+require 'webmock/rspec'
```
h-lame: Could probably put this into spec_helper or rails_helper or a support file so the webmock rspec dsl is available to all our specs.

h-lame:
Looks good!

otlaitil:
:+1: !

PR #64, opened 2015-05-26
Expand README.md to include better instructions for environment set up

alexander-bobin:
I suspect I've missed some railsy things here but this is a start and certainly what got me up and running.


h-lame:
All good.  Thanks @alexander-bobin for keeping us more uptodate.

PR #65, opened 2015-05-26
Reduce number of trending petitions to 3

pixeltrix:
From analysis of previous petition data there isn't really any need to
show up to 12 petitions as trending because only 2-3 are trending at
any particular time.

Also this removes the TrendingPetition model as it's an unnecessary
complication - we can fix any peformance problems with the query by
caching the last verified signature time on the petition and then
loading in the signature count from there. Also this will likely be
cached using 'Russian Doll Caching' when we re-implement the home page.

https://www.pivotaltracker.com/story/show/95404898


otlaitil:
LGTM :+1:

mybits:
LGTM :+1:

h-lame:
Rebased, merging when green.

PR #69, opened 2015-05-27
Don't display validated petitions as trending

pixeltrix:
If a petition has been signed by the sponsors but not yet moderated then
it shouldn't appear in the trending petitions list and report.

https://www.pivotaltracker.com/story/show/95404898

### DIFF
--- a/app/models/petition.rb
+++ b/app/models/petition.rb
@@ -75,6 +75,7 @@ class Petition < ActiveRecord::Base
   scope :moderated, -> { where(state: MODERATED_STATES) }
   scope :trending, ->(number_of_days) {
                       joins(:signatures).
+                      where("petitions.state" => "open").
                       where("signatures.state" => "validated").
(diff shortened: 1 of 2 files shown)


alanth:
LGTM

h-lame:
Should it try to avoid closed petitions too?  "closed" is a calculated state where ``state == 'open' && closed_at < Time.current`` ? I can see that it might be interesting to see that a petition was "trending" as it approached it's closing date, but also maybe disappointing that you can't also sign?  One for @yahoopete ?

yahoopete:
Yeah,  it ideally should include closed because it did actually get the
signs.  You can't actually sign,  but it's what you were looking for,  so
better to let you find it easily and break the news when you select

But only if that's easier... Its pretty edge case. Post midnight,  going
via homepage.

h-lame:
Ok, nothing to do here (it already includes closed petitions).

yahoopete:
:shipit:

PR #70, opened 2015-05-28
Clean up url generation and testing

pixeltrix:
This cleans up a bunch of stuff with how we generate, use and test urls - see individual commit messages for details of each change.


alanth:
No more "secure => true" \o/

PR #72, opened 2015-05-28
Searching scope for pre-create petition query changed

mybits:
from negative positive one - to get only 2 states we want and nothing else


app/decorators/petition_search.rb
```diff
@@ -68,7 +68,7 @@ def execute_search_for_create_petition_query
     Petition.search do |query|
       query.fulltext search_term_sanitised
       query.paginate page: 1, per_page: 3
-      query.without(:state, 'hidden')
+      query.with(:state, ['open', 'rejected'])
```
otlaitil: :+1:

mybits:
Thanks @otlaitil! Merging then.

PR #73, opened 2015-05-29
Dont collect sponsors provide a token url instead

h-lame:
For: https://www.pivotaltracker.com/story/show/94783872

We don't collect the list of sponsors during petition creation, instead we send the creator an email with a link they can send to folk to get them to sign up on their own.  ``Sponsor`` now acts as a join table so we remove the email and token fields from it, and don't create them until after a signature has been created via the form.


h-lame:
Our routes for sponsoring are now:

* ``GET /petitions/<id>/sponsors/<token>`` - shows the form
* ``PATCH /petitions/<id>/sponsors/<token>`` - submit the form and create the sponsor/signature (used to be an update to an existing object - maybe we should change this to POST?)
* ``GET /petitions/<id>/sponsors/<token>/thank-you`` - the "you've almost signed, now check your email" page
* ``GET /petitions/<id>/sponsors/<token>/sponsored`` - the "thanks for validating your email, you've signed as a sponsor" page

The last two follow the pattern from signatures where success in the create action results in a redirect to a page with some info on it.  You could access those pages directly if you had the ids and tokens, but it wouldn't be a problem because they're GETs that don't have side-effects.  My question is do we think we even need the ``<token>`` in there at all?  Could easily just be ``GET /petitions/<id>/sponsors/[thank-you|sponsored]``

app/models/petition.rb
```diff
@@ -24,6 +24,9 @@
 
 class Petition < ActiveRecord::Base
   include State
+  include PerishableTokenGenerator
+
+  has_perishable_token called: 'sponsor_token'
```
otlaitil: This is cool :beers:

otlaitil:
LGTM :+1:

h-lame:
Rebased, maybe merge if it goes :green_apple: so it doesn't linger too long.

PR #74, opened 2015-05-29
Merge accessibility, privacy, cookies with T&Cs

yahoopete:
As per: https://www.pivotaltracker.com/story/show/95695170

This removes the need for a load of separate pages that basically start treading on each others toes, and tidies up the footer in the new designs. 

Separate story to address the content and flow of this page to follow.


h-lame:
If we add ids to the h1's (maybe h2's as well) on the new help page we can link people directly to specific sections.  E.g. when we have ``link_to 'Terms & conditions', help_path`` folk will end up at the top of the page and have to scroll to find the t&cs, we could do ``link_to 'Terms & Conditions', help_path(anchor: 'terms-and-conditions')`` and send them down to the right place.

Also we need a rebase on this.

alexander-bobin:
Would be good to get this merged in prior to the petitions show restyle

h-lame:
Looks good!

h-lame:
Rebased and will merge when :green_book:

PR #75, opened 2015-05-29
Style view petition

alexander-bobin:
This is a big one unfortunately. Pay most attention to the first and last commit.

There's a load of image inclusion and scss copy and pastes which you can ignore.

:8ball:


.gitignore
```diff
@@ -9,6 +9,8 @@ public/cucumber
 public/index.html
 public/javascripts/all.js
 public/api/**
+public/assets/application-*.css
+public/assets/.sprockets-*.json
```
pixeltrix: We should just ignore the whole assets directory

.gitignore
```diff
@@ -19,3 +21,7 @@ coverage
 */.DS_Store
 */*/.DS_Store
 .idea
+Gruntfile.js
+package.json
+node_modules
+.ruby-version
```
pixeltrix: Are we building the assets with Grunt? Why are these here?

pixeltrix:
Why the``@javascript` for email confirmations to work? Isn't it the fact that we need to run the mail delivery job?

.gitignore
alexander-bobin: I am using grunt locally for LiveReload

.gitignore
alexander-bobin: Makes sense. Will change

alexander-bobin:
It is because the steps involve interacting with a `detail` and `summary` node. This is a new HTML5 pattern used by govuk. Murray tells me the `@javascript` causes the use of a real(er?) browser which seems to handle these elements better. It doesn't seem to work without the `@javascript`

otlaitil:
Looks good!

mybits:
LGTM :+1:

PR #76, opened 2015-05-29
Basic healthcheck page

h-lame:
For: https://www.pivotaltracker.com/story/show/95514606

### DIFF
--- a/lib/health_check_middleware.rb
+++ b/lib/health_check_middleware.rb
@@ -1,0 +1,21 @@
+require 'health_check'
+
+class HealthCheckMiddleware
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    if env['PATH_INFO'] =~ %r[\A/health-check/?\Z]
+      [
+        200,
+        {
+          'Content-Type' => 'application/json',
+        },
+        [HealthCheck.checkup(env).to_json]
+      ]
+    else
+      @app.call(env)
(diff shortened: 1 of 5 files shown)


lib/health_check_middleware.rb
```diff
@@ -0,0 +1,21 @@
+require 'health_check'
+
+class HealthCheckMiddleware
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    if env['PATH_INFO'] =~ %r[/health-check/?\Z]
```
pixeltrix: Do we want to anchor this to the beginning of the string as well?

pixeltrix:
I'm worried that some of those Socket methods could be time-consuming - especially ones relating to DNS lookups and that it could end up triggering an autoscale event on the load balancer when it's pinging app servers for status.

Also, I wonder whether we should be returning 5xx errors if things like the DB connection isn't active?

alexander-bobin:
This PR is stagnating. Are we happy to leave it for when Murray gets back?

yahoopete:
@h-lame is back Monday, so we can leave it till then.

pixeltrix:
@h-lame did you see my note about the path regex? It'll match any url ending with `/health-check` currently - is this intentional?

h-lame:
Agh, no. I missed it.  Will fix and reupload.

PR #77, opened 2015-05-31
Implement archived e-petitions

pixeltrix:
Basic implementation of archived e-petitions - will be fleshed out more by later stories.

https://www.pivotaltracker.com/story/show/95512438

### DIFF
--- a/app/controllers/petitions_controller.rb
+++ b/app/controllers/petitions_controller.rb
@@ -12,7 +12,15 @@ class PetitionsController < ApplicationController
   end
 
   def show
-    respond_with @petition = Petition.visible.find(params[:id])
+    begin
+      respond_with @petition = Petition.visible.find(params[:id])
+    rescue ActiveRecord::RecordNotFound => e
+      if @petition = ArchivedPetition.find_by_id(params[:id])
+        redirect_to archived_petition_url(@petition)
+      else
+        raise e
+      end
+    end
   end
 
   def new
--- a/app/models/archived_petition.rb
+++ b/app/models/archived_petition.rb
@@ -10,6 +10,20 @@ class ArchivedPetition < ActiveRecord::Base
   searchable do
     text :title
     text :description
+    time :created_at, trie: true
+  end
+
+  class << self
+    def search(params)
+      query = params[:q].to_s
+      page  = [params[:page].to_i, 1].max
+
+      solr_search do
+        fulltext query
+        paginate page: page, per_page: 20
+        order_by 'created_at', 'asc'
+      end.results
+    end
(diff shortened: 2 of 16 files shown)


pixeltrix:
Hmm, not sure why it's failing - works locally. I'll investigate.

app/models/archived_petition.rb
```diff
@@ -0,0 +1,40 @@
...
+      end.results
+    end
+  end
+
+  def open?
+    state == OPEN_STATE && closed_at.nil?
+  end
+
+  def closed?(time = Time.current)
+    state == OPEN_STATE && !!closed_at && closed_at <= time
```
otlaitil: Nitpick warning: https://github.com/bbatsov/ruby-style-guide#no-bang-bang

app/controllers/petitions_controller.rb
```diff
@@ -12,7 +12,15 @@ def index
   end
 
   def show
-    respond_with @petition = Petition.visible.find(params[:id])
+    begin
```
otlaitil: You could drop the ```begin``` here?

otlaitil:
LGTM :+1:

app/models/archived_petition.rb
```diff
@@ -0,0 +1,40 @@
+class ArchivedPetition < ActiveRecord::Base
+  OPEN_STATE = 'open'
```
otlaitil: Could we use the ```State``` concern?  With new constant ```STATES_ARCHIVED = [OPEN_STATE, REJECTED_STATE]``` or something?

app/models/archived_petition.rb
pixeltrix: @otlaitil normally I'd agree but if `closed_at` is `nil` then the return value of `closed?` becomes `nil` instead of `false`. I'd prefer that we returned a consistent value rather than one that varies.

app/models/archived_petition.rb
pixeltrix: The `State` concern has many states that aren't applicable to archived petitions - using it seemed to complicate things rather than making them simpler.

app/controllers/petitions_controller.rb
pixeltrix: Yes, you could. I prefer explicit blocks because they fold nicer in my text editor - it's purely a subjective decision.

otlaitil: Ok!

app/models/archived_petition.rb
otlaitil: Agreed. :)

PR #78, opened 2015-06-01
Add constituency info to signature signed page

davidbasalla:
For the constituency info to appear on the signed page, I had to enable member
routes for the signed signature to access the signature's postcode. Eager to hear
alternative/better approaches of how to achieve this!

### DIFF
--- a/app/controllers/signatures_controller.rb
+++ b/app/controllers/signatures_controller.rb
@@ -1,6 +1,6 @@
 class SignaturesController < ApplicationController
   before_filter :retrieve_petition, :only => [:new, :create, :thank_you, :signed]
-  before_filter :retrieve_signature, :only => [:signed]
+  before_filter :retrieve_signature, :only => [:signed, :verify, :unsubscribe]
   include ActionView::Helpers::NumberHelper
 
   respond_to :html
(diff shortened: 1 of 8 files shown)


app/controllers/signatures_controller.rb
```diff
@@ -1,5 +1,6 @@
 class SignaturesController < ApplicationController
   before_filter :retrieve_petition, :only => [:new, :create, :thank_you, :signed]
+  before_filter :retrieve_signature, :only => [:signed]
```
otlaitil: You could probably use this new ```before_filter``` in other actions in the controller as well (```verify``` and ```retrieve_signature``` at least)

spec/controllers/signatures_controller_spec.rb
```diff
@@ -5,6 +5,16 @@
     context "signature of user who is not the petition's creator" do
       let(:petition) { FactoryGirl.create(:petition) }
       let(:signature) { FactoryGirl.create(:pending_signature, :petition => petition) }
+      let(:api_url) { "http://data.parliament.uk/membersdataplatform/services/mnis/Constituencies" }
+      let(:fake_body) {
+        "<Constituencies>
```
otlaitil: Just wondering if our super cool ConstituencyApi could have some test helper to generate these mock responses for us & maybe help to stub the requests? Just an idea. 

At least we could use ```let(:api_url) { ConstituencyApi::Client.API_URL }```

otlaitil:
LGTM :+1:

alanth:
LGTM. Merging

PR #80, opened 2015-06-01
Use dotenv gem to manage environment variables

pixeltrix:
Using the dotenv gem allows us to commit the database.yml file and
simplify the Travis CI setup. It also needed because the production
database connection will be available on the DATABASE_URL environment
variable as well as the Rails secret token.


otlaitil:
LGTM :+1:

mybits:
:+1:

PR #81, opened 2015-06-01
Remove email encryption as it's pointless

pixeltrix:
Encrypting the emails complicates the code, slows down the database
queries, uses extra CPU cycles and ultimately more energy.

To address the security arguments for encryption:

1.  It protects people from identification if the database is leaked
    and they've signed contentious or provocative petitions.

    Email encryption doesn't provide this protection as we're
    recording lots of other identifying information like name,
    postcode and ip address.

2.  It protects people from being spammed if the database is leaked.

    Email encryption doesn't provide even this protection because the
    database of signatures shouldn't be downloaded for any other use
    so any likely attack scenario would involve compromise of the
    server and therefore would also compromise the encyption key as
    well. The data would then be trivial to decrypt.


alexander-bobin:
Did I miss a step in here which decrypts emails in the table? Or is that not necessary after archiving in other stories?

alanth:
I approve of this as:

1. It was a response to a "panic attack" of people leaving laptops in public places, e.g. trains

2. It significantly reduces performance

3. We can secure our servers and change keys in an instant if they are compromised (unlikely)

pixeltrix:
@alexander-bobin there is no live data yet so nothing to decrypt :smile:

yahoopete:
I agree that email encryption is pointless by itself - so a question.... if we were to 'encrypt the database' would we treat it in a very different way?  If so, I'd be inclined to merge this puppy.  :dog:

yahoopete:
Noted that we'll be encrypting the whole database now.

PR #82, opened 2015-06-01
Removing duplicate crown copyright link from footer

yahoopete:
Application of new design, now renders the previous crown copyright link redundant.


davidbasalla:
Good read! +1

PR #83, opened 2015-06-02
Quick change to archive show markup to allow petition steps to pass

alexander-bobin:


mybits:
LGTM :palm_tree:

yahoopete:
I'd be better to show something like the below:

"This petition was open between 21 March 2012 and 21 March 2013"

But happy to worry about this in a separate story if this is scope creep with this.....

alexander-bobin:
Let's create a separate story for optimisations. This branch was really just to get the integration tests passing.

yahoopete:
Fair enough.

PR #85, opened 2015-06-02
Add import task for archived petitions

pixeltrix:
This task imports petitions from a file specified by the FILE argument.
It will skip over any petitions not in a open or rejected state and
concatenates the generated rejection reason with any rejection text so
that any future updates to those reasons will not affect historical data.


alanth:
:shipit:

PR #86, opened 2015-06-02
Fix require of Rails spec helper

pixeltrix:
There's no need to do a relative require as the load paths are setup correctly.


davidbasalla:
+1

otlaitil:
:+1:

PR #87, opened 2015-06-02
Start using locales

pixeltrix:
We should start using locales so it seemed like a good time to fix some deprecations as well.


otlaitil:
:+1:

PR #88, opened 2015-06-02
Fix hrefs on favicon links

pixeltrix:
The favicon links need to be root relative otherwise browsers will look
for them in the wrong place, e.g:

```
/archived/petitions/10001
```

will look for a favicon at:

```
/archived/petitions/favicon.ico
```

This will end up causing a database query looking for an archived petition
with an id of 'favicon' and try to render a .ico format template.


alexander-bobin:
LGTM

PR #89, opened 2015-06-02
Remove redundant time format

pixeltrix:
The dotted_short_date format has no dots and is only used in one place
so there's no point adding it to Time::FORMATS.


alexander-bobin:
LGTM

davidbasalla:
+1

PR #90, opened 2015-06-03
Start new petitions with id >= 100000

pixeltrix:
So that the new petition ids don't overlap with the old petition ids, start the petion id sequence from 100000. To do this we need to flatten the migrations and convert to using SQL for the schema format to record the sequence values.

https://www.pivotaltracker.com/story/show/95512706


alanth:
:+1:

otlaitil:
LGTM :+1:

davidbasalla:
cool

PR #92, opened 2015-06-03
Manage sponsorship attempts after reaching limits of 20 sponsors

mybits:
different scenarios described here:
PT: https://www.pivotaltracker.com/n/projects/307301/stories/94792936

Added one commit to fix signature route after rebasing

Added new commit after updating ACs - we want also stop collecting sponsors on petitions that are still pending (no one has validated its email)

### DIFF
--- a/app/views/petitions/moderation_info.html.erb
+++ b/app/views/petitions/moderation_info.html.erb
@@ -1,0 +1,9 @@
+<% content_for(:page_title) { "#{@petition.title} - has been already sent to moderation - e-petitions" } %>
+<% content_for :title do %>
+  <div class="title_block">
+    <h2><%= @petition.creator_signature.name %>'s petition - "<%= @petition.title %>" already has its 5 sponsors and we're taking a look to make sure it meets the Petitions service terms and conditions.<h2>
+    <h2>The petition should be available for signing in a few days.</h2>
+  </div>
+<% end %>
+
+<%= link_to "Return to home page", home_path, :class => 'link_button large_link_button return_to_homepage_button' %>
--- a/spec/controllers/signatures_controller_spec.rb
+++ b/spec/controllers/signatures_controller_spec.rb
@@ -86,6 +86,15 @@ describe SignaturesController do
         end
       end
 
+      it 'doesn\'t send email notification to the petition creator if the petition is already in moderation' do
+        petition_in_moderation = FactoryGirl.create(:petition, state: 'sponsored')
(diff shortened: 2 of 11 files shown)


spec/controllers/signatures_controller_spec.rb
```diff
@@ -86,6 +86,17 @@
         end
       end
 
+      it 'doesn\'t send email notification to the petition creator if the petition is already in moderation' do
+        petition_in_moderation = FactoryGirl.create(:petition, state: 'sponsored')
+        sponsor = FactoryGirl.create(:sponsor, petition: petition_in_moderation)
+        signature = sponsor.create_signature(FactoryGirl.attributes_for(:pending_signature, petition: petition_in_moderation))
+        perform_enqueued_jobs do
```
otlaitil: We could probably use another approach to assert that the request does not add any jobs to the queue:  http://edgeapi.rubyonrails.org/classes/ActiveJob/TestHelper.html#method-i-assert_no_enqueued_jobs
```
assert_no_performed_jobs do
  get :verify, :id => signature.id, :token => signature.perishable_token
end
```

otlaitil:
Looks good! :+1: The test suite is very comprehensive!

spec/controllers/signatures_controller_spec.rb
mybits: Oh that's neat! Will implement!

mybits: Just added it here: https://github.com/alphagov/e-petitions/commit/abc5d4bd5f63e73f9c6f0db5ec2128df5982946a
Thans for the tip!

app/views/petitions/moderation_info.html.erb
```diff
@@ -0,0 +1,9 @@
+<% content_for(:page_title) { "#{@petition.title} - has been already sent to moderation - e-petitions" } %>
+<% content_for :title do %>
+<div class="title_block">
+  <h1>Thank you for your support, but <%= @petition.title %> petition has been already sent to moderation.</h1>
```
yahoopete: Can we change the words to:

"<name of creator>'s petition "<petition title>" already has its 5 sponsors and we're taking a look to make sure it meets the Petitions service terms and conditions."

"The petition should be available for signing in a few days".

mybits: Sure, changed here: https://github.com/alphagov/e-petitions/commit/426c61a16b2db7d441c2bc35ae354fb76cf23915

davidbasalla:
looks good to me!

PR #93, opened 2015-06-03
Icon pack

henryhadlow:
All the icons.

See _icons.scss or _header.html.erb for how to use


otlaitil:
Why are we not using the asset pipeline to serve the icons?

app/views/shared/_os.html.erb
```diff
@@ -0,0 +1,34 @@
+<!-- Desktop Browsers -->
+<link rel="shortcut icon" type="image/x-icon" href="/favicon.ico" />
+
+<!-- Android: Chrome M39 and up-->
+<link rel="manifest" href="/manifest.json">
+
+<!-- Android: Chrome M31 and up, ignored if manifest is present-->
+<meta name="mobile-web-app-capable" content="yes">
+<link rel="icon" sizes="192x192" href="/images/os-social/android/icon-192x192.png">
```
alexander-bobin: Seems like this file does not exist

alexander-bobin:
Maybe we could include our images in the asset pipeline as part of another story?

alexander-bobin:
LGTM

PR #94, opened 2015-06-03
Add field converters to import task

pixeltrix:
The CSV library can take an array of converters that convert the string from the csv field and if any of them return a non-string value then it uses that value. If it doesn't match any of them then it will use the original string value.

The converters are:

```
Null:    If a string matches "NULL" then convert it to nil
Integer: If a string is all digits then convert it to an integer
Time:    If a string matches the database timestamp format then convert it to a UTC time.
````


mybits:
LGTM :+1:

otlaitil:
:+1:

PR #96, opened 2015-06-04
Refactor help

henryhadlow:


yahoopete:
You'll need to remove the how it works page from a bunch of places, not just the page itself!

henryhadlow:
Oh yeah Pete you are right

PR #98, opened 2015-06-04
Add error catching for Faraday errors

davidbasalla:
Added error catching for other Faraday errors other than timeout. Reduced timeout to 5 seconds.
Also removed the constituency info from the petition creation replay stage.


otlaitil:
LGTM :+1:

mybits:
Probably better would be to do it in two separate commits (removing the constituency info) - no need to change it now though, just heads up for the next time:)
LGTM :cake:

PR #102, opened 2015-06-05
Add signature postcode setter method to strip white spaces

davidbasalla:
Allows for user input of postcodes with arbitrary white spaces. The signature 
attribute setter will strip all white spaces and upcase the postcode. As before 
at validation the string will checked against the postcode regex.
Have also removed some postcode tests there were testing different 
postcode formats, which is not needed anymore since the attr setter tests 
ensure that the postcode is squished and upcased.

Addresses https://www.pivotaltracker.com/story/show/95968600

### DIFF
--- a/spec/models/signature_spec.rb
+++ b/spec/models/signature_spec.rb
@@ -20,12 +20,12 @@
 require 'rails_helper'
 
 describe Signature do
-  it "should have a valid factory" do
+  it "has a valid factory" do
     expect(FactoryGirl.build(:signature)).to be_valid
   end
 
   context "defaults" do
-    it "state should default to pending" do
+    it "has pending as default state" do
       s = Signature.new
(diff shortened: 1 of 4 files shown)


mybits:
LGTM :dart:

spec/models/signature_spec.rb
```diff
@@ -52,6 +52,33 @@
...
+      end
+      it "should remove whitespaces and upcase the postcode" do
+        signature.postcode = "   N1  1ty "
+        expect(signature.postcode).to eq "N11TY"
+      end
+    end
+    describe "#email=" do
+      let(:signature) { FactoryGirl.build(:signature) }
+
+      it "should downcase the email" do
```
otlaitil: +1 for adding the missing test case for this existing method!

otlaitil:
LGTM :+1:

PR #104, opened 2015-06-05
Changed home link from e-petitions home to Home in shared footer

yahoopete:
No tests affected, possibly the most challenging bit of hacking you're likely to see this year.


yahoopete:
http://www.nickadamsweb.com/wp-content/uploads/2009/06/white-kid-dance-club.gif

alexander-bobin:
So smug

PR #105, opened 2015-06-06
Deploy process improvements

oskarpearson:
Use per-environment buckets: So that we don't have to create cross-aws-account permission references, the release bucket name should vary, rather than just the directory.

If we install the same version of the code on a system twice, tmp would already exist. This would lead to an error making the directory.


pixeltrix:
LGTM, lets :shipit:

PR #107, opened 2015-06-08
Re-flatten migrations

pixeltrix:
When #81 was merged it added a migration that was generated before the migrations were flattened in #90. This re-flattens the migrations so that a `rake db:migrate` will work with an empty database.


h-lame:
:+1:

PR #108, opened 2015-06-08
Add cookie information message

pixeltrix:
When a user visits a site for the first time show them a message about cookies and link to the cookies section on the help page. On any subsequent page views the cookie message is hidden by the presence of a cookie called `seen_cookie_message` being set to `yes`.

This follows the pattern establised on GOV.UK and in the [govuk_template gem][1] but the implementation is slightly different in that we're not using JavaScript to determine whether to show the message or not and we're also setting the expiry date on the cookie that stores whether the user has seen the message or not to 1 year and not the standard 28 days.

We may need to revisit this when we look at performance and caching.

[https://www.pivotaltracker.com/story/show/95857040][2]

[1]: https://github.com/alphagov/govuk_template
[2]: https://www.pivotaltracker.com/story/show/95857040

### DIFF
--- a/features/joe_sees_cookie_message.feature
+++ b/features/joe_sees_cookie_message.feature
@@ -1,0 +1,19 @@
+Feature: Joe sees cookie message
+  In order to be informed about privacy issues
+  As Joe, a member of the general public
+  I want to see a message about cookie usage
+
+  Scenario: On the first visit
+    Given I am on the home page
+    Then I should see the cookie message
+
+  Scenario: On subsequent visits
+    Given I am on the home page
+    And I go to the home page
+    Then I should not see the cookie message
+
+  Scenario: On revisiting after a year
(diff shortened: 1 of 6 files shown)


h-lame:
:+1: nice and simple.

otlaitil:
LGTM :+1:

features/joe_sees_cookie_message.feature
```diff
@@ -0,0 +1,19 @@
...
+    Then I should see the cookie message
+
+  Scenario: On subsequent visits
+    Given I am on the home page
+    And I go to the home page
+    Then I should not see the cookie message
+
+  Scenario: On revisiting after a year
+    Given I am on the home page
+    And I wait for 1 year
```
alexander-bobin: Nice

alexander-bobin:
LGTM

PR #109, opened 2015-06-09
Petitions in sponsorhip in admin view

mybits:
Admin to be able to filter in todo list page the petitions that are still collecting sponsors. The default list on index page is the list of petitions that await a moderation.
That PR also fix admin being unable to edit or update a petition that is not in the sponsored state - we want admin to be able to edit/update the petition in pending/validated state. For these states the current status will be displayed on edit petting view page (just to show you're about to edit the petition in non-sponsored state).

PT: https://www.pivotaltracker.com/n/projects/307301/stories/95523540


app/models/concerns/state.rb
```diff
@@ -15,4 +15,9 @@ module State
   CLOSED_STATE = 'closed'
   SELECTABLE_STATES = [OPEN_STATE, CLOSED_STATE, REJECTED_STATE, HIDDEN_STATE]
   SEARCHABLE_STATES = [OPEN_STATE, CLOSED_STATE, REJECTED_STATE]
+  TODO_LIST_STATES = [SPONSORED_STATE, VALIDATED_STATE, PENDING_STATE]
+  COLLECTING_SPONSORS_STATES = [VALIDATED_STATE, PENDING_STATE]
+  COLLECTING_SPONSORS = 'collecting sponsors'
```
otlaitil: I am not sure if it is a good idea to mix real states (meaning those persisted in the DB) with these filter keywords. Maybe there could be a better place for these constants? Or they could be named clearly to not be confused with real states. 

Mixing these in the controller and model is a bit confusing as well. Of course this is hard if we want to stick with the single `select` html element (for filtering with a single state and a group of  states).

Just some thoughts, I don't know if any action is needed. What do you think?

app/views/admin/todolist/index.html.erb
```diff
@@ -1,4 +1,12 @@
-<%= render 'admin/shared/results_per_page', :path => admin_root_path %>
+<%= form_tag admin_root_path, method: :get, class: 'filter' do %>
+  <%= radio_button_tag :state, Petition::IN_MODERATION, true %>
+  <%= label_tag :state_awaiting_moderation, 'awaiting moderation', class: 'radio-button' %>
+  <%= radio_button_tag :state, Petition::COLLECTING_SPONSORS,  params[:state] == 'collecting sponsors' ? true : false %>
```
otlaitil: Nitpick: you can drop the ` ? true : false ` part :)

mybits: True, thanks :)

app/models/concerns/state.rb
h-lame: The ``COLLECTING_SPONSORS``, ``TODO_LIST``, and ``IN_MODERATION`` constants are only here to allow the ``.for_state`` scope to work.  We could be more careful about when we use ``for_state`` and prefer the ``in_moderation``, ``collecting_sponsors``, and ``todo_list`` scopes instead and that way we could drop these constants entirely.

mybits:
@h-lame and @otlaitil thanks so much for your comments! All addressed in relevant commits.

app/models/concerns/state.rb
mybits: Oh somehow I've missed that comment, so not ready for re-review yet, sorry

mybits: @h-lame & @otlaitil now's ready, sorry for false start and thanks for the comments!

app/controllers/admin/todolist_controller.rb
```diff
@@ -1,6 +1,10 @@
 class Admin::TodolistController < Admin::AdminController
-
   def index
-    @petitions = Petition.for_state(Petition::SPONSORED_STATE).order(:created_at).paginate(:page => params[:page], :per_page => params[:per_page] || 20)
+    @petitions = Petition.in_moderation.order(:created_at)
+    if params[:petition] == 'collecting_sponsors'
+      @petitions = Petition.collecting_sponsors.order(:created_at)
+    end
+    @petitions = @petitions.paginate(page: params[:page], per_page: params[:per_page] || 20)
```
h-lame: Nitpick:  This might be clearer as:

```ruby
if params[:petition] == 'collecting_sponsors'
  @petitions = Petition.collecting_sponsors
else
  @petitions = Petition.in_moderation
end
@petitions = @petitions.order(:created_at).paginate(page: params[:page], per_page: params[:per_page] || 20)
```

There might even be a cleaner way to express it than that (not a huge fan of ``@petitions = @petitions.<blah>`` at the end.

mybits: Yes, much cleaner, thanks! Changes and removing redundant instant variables here:https://github.com/alphagov/e-petitions/commit/f8cde133388029058ff8bafe0cf7d34ef305a8a1

app/controllers/admin/todolist_controller.rb
```diff
@@ -1,6 +1,11 @@
 class Admin::TodolistController < Admin::AdminController
-
   def index
-    @petitions = Petition.for_state(Petition::SPONSORED_STATE).order(:created_at).paginate(:page => params[:page], :per_page => params[:per_page] || 20)
+    if params[:petition] == 'collecting_sponsors'
+      petitions = Petition.collecting_sponsors
+    else
+      petitions = Petition.in_moderation
+    end
+    @petitions = petitions.order(:created_at).paginate(page: params[:page], per_page: params[:per_page] || 20)
```
pixeltrix: I'd move this logic to a scope on the model - keeping the `paginate` method in the controller is okay though.

mybits: Yes that's better! Thanks @pixeltrix, changed here: https://github.com/alphagov/e-petitions/commit/e83ad6e3557d5650db468135ab43a1d4e7fc515a

PR #110, opened 2015-06-09
Redirect sponsors of already open petitions, to the view page

yahoopete:
This is for story > https://www.pivotaltracker.com/story/show/96533342

It's a follow up to the already delivered story, https://www.pivotaltracker.com/story/show/94792936 where we manage sponsorship attempts in various ways when we don't need / want any more.

The story should have been directing people to the petition view page not the petition sign page (I used the wrong name when I wrote the story....)

This PR remedies that....


h-lame:
seems legit. :+1:

mybits:
LGTM :ship:

yahoopete:
raise_the_roof_newsman-298x157

PR #111, opened 2015-06-09
Remove old API

pixeltrix:
The old API isn't fit for purpose so let's remove it until we can write something better.


yahoopete:
Yep - do it.*

*this is not a code review

h-lame:
Could remove rabl from the gemfile, if we do an API later we can decide what tech to use for it at that point.

PR #114, opened 2015-06-09
Add MP info and URL to constituency info for signed signature

davidbasalla:
for https://www.pivotaltracker.com/story/show/95870752

### DIFF
--- a/app/lib/constituency_api.rb
+++ b/app/lib/constituency_api.rb
@@ -1,25 +1,32 @@
 module ConstituencyApi
   ConstituencyApiError = Class.new(RuntimeError)
 
-  class Constituency < Struct.new(:name)
+  class Constituency < Struct.new(:name, :mp)
   end
-  
+
+  class Mp < Struct.new(:id, :name, :start_date)
+    URL = "http://www.parliament.uk/biographies/commons"
+    def url
+      "#{URL}/#{name.parameterize}/#{id}"
+    end
+  end
+
   class Client
     include Faraday
     URL = 'http://data.parliament.uk/membersdataplatform/services/mnis/Constituencies'
     TIMEOUT = 5
-    
+
     def self.constituencies(postcode)
       response = call_api(postcode)
       parse_constituencies(response)
     end
-    
+
     def self.parse_constituencies(response)
       return [] unless response["Constituencies"]
       constituencies = response["Constituencies"]["Constituency"]
-      Array.wrap(constituencies).map { |c| Constituency.new(c["Name"]) }
+      Array.wrap(constituencies).map { |c| Constituency.new(c["Name"], last_mp(c)) }
     end
-    
+
     def self.call_api(postcode)
       response = Faraday.new(URL).get("#{postcode_param(postcode)}/") do |req|
         req.options[:timeout] = TIMEOUT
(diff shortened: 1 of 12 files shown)


app/views/signatures/signed.html.erb
```diff
@@ -16,7 +16,11 @@
 <% if @signature.constituency %>
   <div class="section-panel">
     <p>
-      Your constituency is <%= @signature.constituency.name %>
+      Your constituency is <%= @signature.constituency.name %><br/>
+      <% if @signature.constituency.mp %>
+        Your local MP is <%= @signature.constituency.mp.name %><br/>
+        <%= link_to "Contact your MP", @signature.constituency.mp.url %>
+      <% end %>
```
h-lame: Minor nitpick: why not do these as ``<p>`` tags rather than using ``<br/>``?

features/step_definitions/petition_steps.rb
```diff
@@ -210,6 +210,22 @@
...
+                                 petition_id: @petition.id)
+  expect(page).to have_text(@signature.constituency.mp.name)
+end
+
+Then /^I can click on a link to visit my MP$/ do
+  @signature = Signature.find_by(email: "womboidian@wimbledon.com",
+                                 postcode: "N11TY",
+                                 name: "Womboid Wibbledon",
+                                 petition_id: @petition.id)
+  expect(page).to have_css("a[href*='#{@signature.constituency.mp.url}']")
```
h-lame: We don't need to use ``@signature`` here as it means the variable will "leak" out of the step and might (possibly) interact unexpectedly with other steps that do use ``@signature``.  If the step can be self-contained it's better, so just use a local variable ``signature`` instead of an instance variable ``@signature`` (instance variables in the context of cucumber are effectively global).

spec/fixtures/constituency_api/SW149RQ.xml
```diff
@@ -0,0 +1 @@
+<Constituencies/>
```
h-lame: It might be clearer if this fixture was called "no_results.xml" rather than named after a specific postcode.

app/lib/constituency_api.rb
```diff
@@ -1,7 +1,14 @@
...
   ConstituencyApiError = Class.new(RuntimeError)
 
-  class Constituency < Struct.new(:name)
+  class Constituency < Struct.new(:name, :mp)
+  end
+
+  class Mp < Struct.new(:id, :name, :start_date)
+    URL = "http://www.parliament.uk/biographies/commons"
+    def url
+      "#{URL}/#{name.gsub(/\s+/, "-")}/#{id}"
```
otlaitil: I think we should be more restrictive when constructing the URL, substituting whitespace might not be enough and we might end up having invalid URLs (if MP's name includes UTF8 characters).

We could probably use this: http://api.rubyonrails.org/classes/ActiveSupport/Inflector.html#method-i-parameterize

davidbasalla:
Thanks all for the comments. Have addressed them and rebased!

app/lib/constituency_api.rb
```diff
@@ -1,25 +1,32 @@
 module ConstituencyApi
   ConstituencyApiError = Class.new(RuntimeError)
```
pixeltrix: why define it this way instead of the more idiomatic:

``` ruby
module ConstituencyApi
  class ConstituencyApiError < RuntimeError; end
end
```

also you could lose the prefix as well:

``` ruby
module ConstituencyApi
  class Error < RuntimeError; end
end
```

then any external rescues can use the shorter `ConstituencyApi::Error`.

app/lib/constituency_api.rb
```diff
@@ -1,25 +1,32 @@
...
   end
-  
+
+  class Mp < Struct.new(:id, :name, :start_date)
+    URL = "http://www.parliament.uk/biographies/commons"
+    def url
+      "#{URL}/#{name.parameterize}/#{id}"
+    end
+  end
+
```
pixeltrix: Are these meant to be value objects? If so `Struct` may not be the best thing to inherit from as it generates `attribute=` methods as well. Also any missing initialize arguments are just nils.

A more explicit way of creating a value object would be something like this:

``` ruby
class Constituency
  attr_reader :name, :mp

  def initlalize(name, mp)
    @name, @mp = name, mp
  end

  def ==(other)
    other.is_a?(self.class) && name == other.name && mp == other.mp
  end
  alias_method :eql?, :==
end
```

This is for future guidance though - it's probably not worth making the change

app/lib/constituency_api.rb
davidbasalla: Nice, much neater! Have fixed and committed.

PR #117, opened 2015-06-10
Remove chronic and timecop gems

pixeltrix:
Our use of these two gems is very limited and can easily be handled using the built-in time helpers, String#in_time_zone and be_within matcher.


mybits:
:+1:

davidbasalla:
:+1:

PR #118, opened 2015-06-10
Update rejected petitions list view

otlaitil:
* Changed default sort ordering
* Disabled showing signature count
* Updated view to match new designs

### DIFF
--- a/app/assets/stylesheets/petitions/views/_shared.scss
+++ b/app/assets/stylesheets/petitions/views/_shared.scss
@@ -101,6 +101,10 @@
   }
 }
 
+.petition-item.petition-rejected {
+  margin-bottom: $gutter;
+}
+
 // About item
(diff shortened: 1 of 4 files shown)


otlaitil:
Addressed @h-lame's PR comments.

app/assets/stylesheets/petitions/views/_shared.scss
```diff
@@ -91,6 +91,10 @@
   }
 }
 
+.petition-item.petition-rejected {
```
alexander-bobin: Nice variation combo. Leave it as is but in future, a variation should have the base in the name so something like `.petition-item-rejected` would be better

alexander-bobin:
LGTM :+1:

davidbasalla:
:shipit:

PR #119, opened 2015-06-10
Remove rspec-collection-matchers gem

pixeltrix:
We're only using the errors_on collection matcher and we can easily write our specs slightly differently to not need it.


h-lame:
Yup.

PR #120, opened 2015-06-10
Remove annotate gem

pixeltrix:
The db/structure.sql file is the source of truth - the annotations always seem to end up out of date any way.


h-lame:
So much yes.

PR #126, opened 2015-06-10
Refactor rejection reasons

pixeltrix:
This does two things:

1.  Uses a locale YAML file for storing the title and descriptions of rejection reasons. Not only is this better because YAML is a human editable format as opposed to JSON it also allows for the possiblity of translating them at some point in the future.

2.  Moves the logic of rejection into the model and out of the controller. This is domain logic and should live in the model.


h-lame:
This seems like a good thing to have done :+1:

spec/factories.rb
```diff
@@ -95,7 +95,7 @@
 
   factory :rejected_petition, :parent => :petition do
     state  Petition::REJECTED_STATE
-    rejection_code "Just do not like"
+    rejection_code "duplicate"
```
otlaitil: :+1: for changing this to a valid rejection_code

otlaitil:
LGTM :+1:

PR #127, opened 2015-06-10
Fix 404 for revalidating signature email

davidbasalla:
The previous retrieve_petition method that was called before the signed controller method was scoped to only OPEN and REJECTED petitions, so would not find any other petition.

### DIFF
--- a/app/controllers/signatures_controller.rb
+++ b/app/controllers/signatures_controller.rb
@@ -1,8 +1,8 @@
 class SignaturesController < ApplicationController
   include ManagingMoveParameter
 
-  before_filter :retrieve_petition, :only => [:new, :create, :thank_you, :signed]
-  before_filter :retrieve_signature, :only => [:signed, :verify, :unsubscribe]
+  before_filter :retrieve_petition, :only => [:new, :create, :thank_you]
+  before_filter :retrieve_signature, :only => [:verify, :unsubscribe]
 
   respond_to :html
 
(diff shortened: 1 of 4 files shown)


jnormington:
QA'd - good to merge when reviewed

app/controllers/signatures_controller.rb
```diff
@@ -19,6 +19,10 @@ def create
     end
   end
 
+  def signed
+    @petition = Petition.find(params[:petition_id])
```
otlaitil: This method currently opens up all petitions from our system and it can be used to show a combination of any petition with any signature, since no checking that they relate is made. This is probably not very good thing.

The route is `petitions/:petition_id/signatures/:signature_id/signed`

So we can just do: 
`http://localhost:3000/petitions/100000/signatures/1/signed`

This is not a very big problem atm, since the page does not reveal any details of the signer. But if it did.. you could spread a link that is saying that someone signed a petition (even rejected one) they did not sign. We should not allow it.

h-lame: What if we moved the ``signed`` action to ``/petitions/<id>/signed`` ?  The signature isn't important to the page we show.   Or we should do as otto suggests and make sure the ``<petition_id>`` and ``<signature_id>`` match up.

davidbasalla: I think we'd need to keep it in signature as we use the signature's constituency method? Will go with what Otto suggested

davidbasalla:
Thanks for the comments and help, have recommitted this with a different approach, by just rendering the signed view from the verify controller method

h-lame:
Seems fine to me, but it might be worth floating this new approach past @yahoopete - I know there were some concerns about not being able to track user journeys properly in GA unless the urls actually changed.

davidbasalla:
Rebased. Still need to check with @yahoopete about new routing approach

alexander-bobin:
@davidbasalla, do you need any help with this PR?

davidbasalla:
Hey, I pushed this up last night. Please have a look and see what you think. Has the revised routing approach where the verify action redirects to an URL such as /petitions/1/signatures/#{signature.perishable_token}/signed. This is to ensure that noone can enter guessable signature id's and see infos about the signature's constituency

alexander-bobin:
@yahoopete, are you ok with this new approach? Seems fine. Murray had a question re: GA above

yahoopete:
If this provides the following URL structure (which it looks to) "https://petition.parliament.uk/petitions/#{signature.petition_id}/signatures/#{signature.perishable_token}/signed" then we are gold....

We can do a thing in GA, which aggregates and anonymises bits of the path (not that it needs anonymising, but it does need aggregating, so we'd see something like 

/petitions/100123/<X>/signed

This will then let me keep tracking journeys etc in GA.

So :+1:

PR #128, opened 2015-06-10
Popular in your constituency

h-lame:
For: https://www.pivotaltracker.com/story/show/95513536

Touches so many things - might make sense to go through commit by commit as while there are alot, they are mostly quite small.

We:

1) store the constituency_id on the signature when it is created - this means we do a lookup for the constituency based on the postcode - we might want to offline this?
2) have a constituency petition journal model where we record the count of signatures for a petition for a given constituency and we increment the count on this whenever a signature is validated (it's silently cool with it if the signature has no constituency_id)
3) allow a user to search by their postcode to find petitions that are popular in their constituency - this throws away petitions that have no signatures (absence of a journal, or a journal with 0 count) for that constituency and orders

### DIFF
--- a/app/controllers/local_petitions_controller.rb
+++ b/app/controllers/local_petitions_controller.rb
@@ -11,6 +11,7 @@ class LocalPetitionsController < ApplicationController
         constituencies = ConstituencyApi::Client.constituencies(sanitized_postcode)
         if constituencies.any?
           @constituency = constituencies.first
+          @petitions = Petition.popular_in_constituency(@constituency.id, 3)
         else
           render 'constituency_lookup_failed'
         end
--- a/app/controllers/local_petitions_controller.rb
+++ b/app/controllers/local_petitions_controller.rb
@@ -1,2 +1,6 @@
 class LocalPetitionsController < ApplicationController
+  respond_to :html
+
+  def index
+  end
 end
--- a/app/controllers/local_petitions_controller.rb
+++ b/app/controllers/local_petitions_controller.rb
@@ -1,6 +1,29 @@
+require 'postcode_sanitizer'
+
 class LocalPetitionsController < ApplicationController
   respond_to :html
 
   def index
+    if sanitized_postcode.blank?
+      render 'no_postcode_provided'
+    else
+      begin
+        constituencies = ConstituencyApi::Client.constituencies(sanitized_postcode)
+        if constituencies.any?
+          @constituency = constituencies.first
+        else
+          render 'constituency_lookup_failed'
+        end
+      rescue ConstituencyApi::Error => e
+        Rails.logger.error("Failed to fetch constituency - #{e}")
+        render 'constituency_lookup_failed'
+      end
+    end
+  end
+
+  private
+
+  def sanitized_postcode
+    PostcodeSanitizer.call(params[:postcode])
   end
 end
--- a/app/controllers/local_petitions_controller.rb
+++ b/app/controllers/local_petitions_controller.rb
@@ -1,0 +1,2 @@
+class LocalPetitionsController < ApplicationController
+end
--- a/app/views/local_petitions/_search.html.erb
+++ b/app/views/local_petitions/_search.html.erb
@@ -1,0 +1,7 @@
+<%= form_tag local_petitions_path, method: 'get' do %>
+  <div class="form-group">
+    <%= label_tag 'postcode', 'UK postcode', class: 'form-label' %>
+    <%= search_field_tag 'postcode', local_assigns[:postcode], class: 'form-control' %>
+  </div>
+  <%= submit_tag 'Search', class: 'button' %>
+<% end %>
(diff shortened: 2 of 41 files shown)


app/controllers/local_petitions_controller.rb
```diff
@@ -0,0 +1,30 @@
+require 'postcode_sanitizer'
+
+class LocalPetitionsController < ApplicationController
+  respond_to :html
+
+  def index
```
otlaitil: What would you think of an approach for another action doing the postcode check and redirecting to a `show` action here.

We could then even leverage friendly id's and have something like `petitions/local/westminster-north` which would be a meaningful URL someone could share.

app/controllers/local_petitions_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+    if sanitized_postcode.blank?
+      render 'no_postcode_provided'
+    else
+      begin
+        constituencies = ConstituencyApi::Client.constituencies(sanitized_postcode)
+        if constituencies.any?
+          @constituency = constituencies.first
+          @petitions = Petition.popular_in_constituency(@constituency.id, 3)
```
otlaitil: Any point making this limit configurable via `AppConfig`?

lib/postcode_sanitizer.rb
```diff
@@ -0,0 +1,5 @@
+module PostcodeSanitizer
+  def self.call(postcode)
+    postcode.to_s.gsub(/\s+/, "").upcase
```
otlaitil: :+1: for extracting this!

spec/support/constituency_api_helpers.rb
```diff
@@ -0,0 +1,79 @@
+require 'postcode_sanitizer'
+
+module ConstituencyApiHelpers
+  module ApiLevel
```
otlaitil: This API helper is nice!

otlaitil:
LGTM. :+1:

app/controllers/local_petitions_controller.rb
h-lame: Good question: @yahoopete ?

app/views/local_petitions/constituency_lookup_failed.html.erb
```diff
@@ -0,0 +1,4 @@
+<h1>Sorry!</h1>
```
alexander-bobin: If this is a page on it's own you could add a class of page-title to the h1 to hook into some existing styles.

app/views/local_petitions/index.html.erb
```diff
@@ -0,0 +1,20 @@
...
+          <%= link_to petition.title, petition_path(petition) %>
+          <p>
+            <%= number_with_delimiter petition.constituency_signature_count %> signatures from your constituency
+            (<%= number_with_delimiter petition.signature_count %> signatures total)
+          </p>
+        </li>
+      <% end -%>
+    </ol>
+  <% else %>
+    <p class="no_petitions_warning">No petitions are popular in your constituency.</p>
```
alexander-bobin: nitpick: This class can go. Seems like an old one.

app/views/local_petitions/_search.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= form_tag local_petitions_path, method: 'get' do %>
+  <div class="form-group">
```
alexander-bobin: :+1: reusing existing GDS classes

app/controllers/local_petitions_controller.rb
h-lame: That's an interesting idea, but we don't have a name -> id lookup for constituencies.  I guess it could be ``/petitions/local/westminister-north-1234/`` or something, but that leads us down the path of having urls where only part of the key is important and people can post urls like ``/petitions/local/look-at-what-these-idiots-care-about-1234/``  which seems bad.

app/views/local_petitions/index.html.erb
h-lame: nope, just copied it from elsewhere (possibly an old elsewhere that doesn't exist) - will remove.

app/views/local_petitions/constituency_lookup_failed.html.erb
h-lame: Cool, will do!

app/controllers/local_petitions_controller.rb
otlaitil: Ah that's true, I was somehow thinking we are persisting also the constituency name to our db.

h-lame:
rebased and addressed @alexander-bobin's styling comments.

lib/postcode_sanitizer.rb
davidbasalla: agreed

h-lame:
Don't merge this yet.  Reminded by #131 that we need to show the MP on the results page and I need to add that.

h-lame:
Added commit with mp info to finish AC.  Now done!

h-lame:
Rebased, plz merge if happy and green.

yahoopete:
kermit-the-muppets1

You talkin to me?

PR #129, opened 2015-06-10
Refactor petition states

pixeltrix:
Since the State module isn't shared by any other module there's no point to it being a concern, so move the constants back to the Petition model.


otlaitil:
LGTM :+1:

h-lame:
:+1: but rebase needed

pixeltrix:
Rebased - once it's green feel free to merge

PR #130, opened 2015-06-11
Switch to PostgreSQL full-text search indexes from Solr

pixeltrix:
This makes our stack simpler, cheaper and easier to deploy.


app/models/concerns/browseable.rb
```diff
@@ -0,0 +1,157 @@
...
+      @page_size ||= [[params.fetch(:count, 50).to_i, 50].min, 1].max
+    end
+
+    def scope
+      @scope ||= facets.keys.detect(-> { :all }){ |key| key.to_s == params[:state] }
+    end
+
+    def scoped?
+      scope != :all
+    end
```
h-lame: Not sure we need this anymore, now that we always have a scope and don't need to branch in execute_query based on scoped or not.

pixeltrix: Possibly, but we may need it in a view so I left it in for now

app/models/petition.rb
```diff
@@ -28,19 +30,13 @@ class Petition < ActiveRecord::Base
...
-    time :created_at, :trie => true
-    string :state
-  end
+  extend Searchable(:title, :action, :description)
+  include Browseable
+
+  facet :all, -> { self.order(:signature_count).reverse_order }
+  facet :open, -> { self.for_state('open').order(:signature_count).reverse_order }
+  facet :closed, -> { self.for_state('closed').order(:signature_count).reverse_order }
+  facet :rejected, -> { self.for_state('rejected').order(:created_at).reverse_order }
```
h-lame: Should these be ``.reorder`` rather than ``.order``?  ``Browseable::Search#execute_search`` calls ``.reorder(:created_at)`` and then instance_execs the facet from params[:state] ontop, so we'd end up with ``ORDER BY created_at, signature_count desc`` right, when I think we'd expect ``ORDER BY signature_count desc``

app/models/concerns/browseable.rb
```diff
@@ -0,0 +1,157 @@
...
+    def query
+      @query ||= params[:q].to_s
+    end
+
+    def page_size
+      @page_size ||= [[params.fetch(:count, 50).to_i, 50].min, 1].max
+    end
+
+    def scope
+      @scope ||= facets.keys.detect(-> { :all }){ |key| key.to_s == params[:state] }
```
h-lame: I don't think it makes much difference, but could this be ``klass.facets.keys`` for the same effect? As I read it ``facets.keys`` and ``klass.facets.keys`` should end up being the same (``facets`` just takes a roundabout way of constructing that set)

features/step_definitions/petition_steps.rb
```diff
@@ -162,7 +162,7 @@
 end
 
 Then /^I should be asked to search for a new petition$/ do
-  expect(page).to have_css("form#pre_creation_search input#search")
+  expect(page).to have_css("form#pre_creation_search input#q")
```
otlaitil: how about just `expect(page).to have_css("#pre_creation_search")`?

app/models/concerns/browseable.rb
```diff
@@ -0,0 +1,157 @@
...
+    class_attribute :facets
+    self.facets = {}
+  end
+
+  class Facets
+    include Enumerable
+
+    attr_reader :search, :klass
+
+    def initialize(search)
```
h-lame: This could take ``klass`` instead of ``search``, we don't seem to use ``search`` other than to call ``.klass`` on it.

spec/models/archived_petition_spec.rb
```diff
@@ -25,7 +21,7 @@
     end
 
     it "sorts the results by the created_at timestamp" do
-      expect(ArchivedPetition.search(q: "Petition")).to eq([petition_2, petition_1])
+      expect(ArchivedPetition.search(q: "Petition")).to match_array([petition_2, petition_1])
```
otlaitil: Isn't the order of the elements in the array important, why change to `match_array`?

app/models/petition.rb
pixeltrix: Actually, I think I need to remove the reorder(:created_at) - good catch

app/models/petition.rb
```diff
@@ -28,19 +30,13 @@ class Petition < ActiveRecord::Base
...
-    integer :signature_count
-    time :closed_at, :trie => true
-    time :created_at, :trie => true
-    string :state
-  end
+  extend Searchable(:title, :action, :description)
+  include Browseable
+
+  facet :all, -> { self.order(:signature_count).reverse_order }
+  facet :open, -> { self.for_state('open').order(:signature_count).reverse_order }
```
otlaitil: Should these use the constants too? ie. `self.for_state(OPEN_STATE)`

otlaitil:
LGTM! :+1: The new arbitrary facet thing is awesome.

app/models/concerns/browseable.rb
```diff
@@ -0,0 +1,157 @@
...
+        end
+
+        hash[key] = scope(key).count
+      end
+    end
+
+    def scope(key)
+      search.klass.instance_exec(&facet(key))
+    end
+  end
```
h-lame: There's a lot of things called ``/facets?/`` in here.  Some that belong to this class, but some that belong to the ``klass``. I wonder if we could rename or delegate to make it clearer what each ``facet`` is.  How does this strike you?

```ruby
class Facets
  include Enumerable

  attr_reader :klass

  def initialize(klass)
    @klass = @klass
  end

  def [](key)
    facet_counts[key]
  end

  delegate :key?, :has_key?, :keys, to: :facet_definitions

  def each(&block)
    keys.each do |key|
      yield key, self[key]
    end
  end

  private
  
  def facet_definitions
    klass.facets
  end

  def facet_counts
    @facets ||= Hash.new(&count_facet_query)
  end

  def count_facet_query
    lambda do |hash, key|
      unless key?(key)
        raise ArgumentError, "Unsupported facet: #{key.inspect}"
      end

      hash[key] = facet_scope(key).count
    end
  end

  def facet_scope(key)
    klass.instance_exec(&facet_definitions[key])
  end
end
```

Not that I'm super happy that all these methods are now ``facet_`` prefixed - it's probably redundant, but leaving them as ``definitions`` or ``counts`` doesn't seem enough...

spec/models/archived_petition_spec.rb
pixeltrix: Yes, it is, unfortunately I wasn't aware of that - I was using it to force `to_ary` to be called. I'll just have to change it back and call `to_a` manually.

app/models/petition.rb
pixeltrix: Probably :-)

app/models/concerns/browseable.rb
pixeltrix: :+1: there is a small performance penalty for using delegate but not worth worrying about

features/step_definitions/petition_steps.rb
pixeltrix: I think I'd rather we took out the ids and just just attribute selectors, e.g:

``` ruby
expect(page).to have_css("form input[name=q]")
```

pixeltrix:
If this is goes green then :shipit:

pixeltrix:
Merging this because I need it for: https://www.pivotaltracker.com/story/show/96278100

PR #131, opened 2015-06-11
Add mp url to constituency info

davidbasalla:
Addressing comments from testing. Revised the HTML layout and added error handling for missing or invalid start_date values for MPs.

### DIFF
--- a/app/lib/constituency_api.rb
+++ b/app/lib/constituency_api.rb
@@ -1,14 +1,35 @@
 module ConstituencyApi
   class Error < RuntimeError; end
 
-  class Constituency < Struct.new(:name, :mp)
+  class Constituency
+    attr_reader :name, :mp
+
+    def initialize(name, mp = nil)
+      @name, @mp = name, mp
+    end
+
+    def ==(other)
+      other.is_a?(self.class) && name == other.name && mp == other.mp
+    end
+    alias_method :eql?, :==
   end
 
-  class Mp < Struct.new(:id, :name, :start_date)
+  class Mp
+    attr_reader :id, :name, :start_date
     URL = "http://www.parliament.uk/biographies/commons"
+
+    def initialize(id, name, start_date)
+      @id, @name, @start_date = id, name, start_date.to_date
+    end
+
     def url
       "#{URL}/#{name.parameterize}/#{id}"
     end
+
+    def ==(other)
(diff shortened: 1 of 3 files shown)


app/lib/constituency_api.rb
```diff
@@ -6,9 +6,25 @@ class Constituency < Struct.new(:name, :mp)
 
   class Mp < Struct.new(:id, :name, :start_date)
     URL = "http://www.parliament.uk/biographies/commons"
+
+    def initialize(id, name, start_date)
+      self.id = id.to_i
```
h-lame: I'm not sure we need to bother with converting this to an integer.  (I know we were before, but it's not super useful.)

app/lib/constituency_api.rb
```diff
@@ -6,9 +6,25 @@ class Constituency < Struct.new(:name, :mp)
 
   class Mp < Struct.new(:id, :name, :start_date)
```
h-lame: Now we've overridden ``initialize`` there's not much we're getting from ``Struct`` - might be time to drop it and fill in the attr / eql stuff ourselves (as @pixeltrix suggested on the previous PR).

h-lame:
Minor comments about the implementation, but it's really good to see the extra test coverage here.

app/lib/constituency_api.rb
```diff
@@ -6,9 +6,25 @@ class Constituency < Struct.new(:name, :mp)
...
     URL = "http://www.parliament.uk/biographies/commons"
+
+    def initialize(id, name, start_date)
+      self.id = id.to_i
+      self.name = name
+      self.start_date = parse_date(start_date)
```
pixeltrix: Is the start_date ever nil? Or anything else? Seems to me that an MP must have a start date otherwise they'd wouldn't be in the list. If it blows up with an ArgumentError and we get an Airbrake notification then that'd be a good indicator that's something's changed about the API - rescuing it and setting it to nil seems likely to mask possible issues.

davidbasalla: I haven't seen any cases where it was nil or a non-standard date string, but Jon suggested to try catching any unexpected errors due to it being an external dataset. 
Would you suggest just to leave it for now then and assume that .to_date() always works?

h-lame:
Looks good, will merge when G R E E N

PR #134, opened 2015-06-11
Use locale keys to correctly pluralize signature counts

pixeltrix:
The i18n gem has a feature where if you pass it an option called count then it will use the sub key called one if the value of count is 1 and if the value of count is more than 1 then it uses other. Coupled with the use of the html key to mark the output as HTML-safe allows us to localize the language and formatting if needed at some point.

https://www.pivotaltracker.com/story/show/96644444

### DIFF
--- a/app/helpers/signature_helper.rb
+++ b/app/helpers/signature_helper.rb
@@ -6,6 +6,10 @@ module SignatureHelper
     end
   end
 
+  def signature_count(key, count)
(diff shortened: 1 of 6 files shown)


pixeltrix:
The spec feels a little like gold-plating but it only took 5 minutes to write so I though I'd add it anyway.

h-lame:
Gonna merge this so I can rebase #128 and use this technique where it writes some signature counts to the page.

PR #136, opened 2015-06-12
Petitions with government response listview

otlaitil:
For https://www.pivotaltracker.com/n/projects/307301/stories/96278298

### DIFF
--- a/app/models/petition.rb
+++ b/app/models/petition.rb
@@ -28,7 +28,7 @@ class Petition < ActiveRecord::Base
 
   has_perishable_token called: 'sponsor_token'
 
-  before_save :stamp_parliament_response_at, if: -> { response.present? && response_was.nil? }
+  before_save :stamp_parliament_response_at, if: -> { response_summary.present? && response.present? && parliament_response_at.nil? }
   after_create :set_petition_on_creator_signature
 
   extend Searchable(:title, :action, :description)
@@ -101,7 +101,7 @@ class Petition < ActiveRecord::Base
 
   scope :by_oldest, -> { order(created_at: :asc) }
 
-  scope :with_response, -> { where.not(response: nil) }
+  scope :with_response, -> { where.not(response_summary: nil, response: nil) }
 
   def self.update_all_signature_counts
     Petition.visible.each do |petition|
--- a/app/models/petition.rb
+++ b/app/models/petition.rb
@@ -38,6 +38,8 @@ class Petition < ActiveRecord::Base
   facet :closed, -> { for_state(CLOSED_STATE).reorder(signature_count: :desc) }
   facet :rejected, -> { for_state(REJECTED_STATE).reorder(created_at: :desc) }
 
+  facet :with_response, -> { where(state: OPEN_STATE).with_response.reorder(parliament_response_at: :desc) }
+
   # = Relationships =
   belongs_to :creator_signature, :class_name => 'Signature'
   accepts_nested_attributes_for :creator_signature
@@ -97,6 +99,8 @@ class Petition < ActiveRecord::Base
                             }
 
   scope :by_oldest, -> { order(created_at: :asc) }
+
+  scope :with_response, -> { where.not(response: nil) }
 
   def self.update_all_signature_counts
     Petition.visible.each do |petition|
--- a/app/models/petition.rb
+++ b/app/models/petition.rb
@@ -28,6 +28,7 @@ class Petition < ActiveRecord::Base
 
   has_perishable_token called: 'sponsor_token'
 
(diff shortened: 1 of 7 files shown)


h-lame:
:+1:

app/models/petition.rb
```diff
@@ -38,6 +39,8 @@ class Petition < ActiveRecord::Base
   facet :closed, -> { for_state(CLOSED_STATE).reorder(signature_count: :desc) }
   facet :rejected, -> { for_state(REJECTED_STATE).reorder(:created_at) }
 
+  facet :with_response, -> { where(state: [OPEN_STATE, CLOSED_STATE]).with_response.reorder(gov_response_at: :desc) }
```
pixeltrix: `CLOSED_STATE` doesn't exist in the database

db/migrate/20150612104023_add_gov_response_at_to_petitions.rb
```diff
@@ -0,0 +1,5 @@
+class AddGovResponseAtToPetitions < ActiveRecord::Migration
+  def change
+    add_column :petitions, :gov_response_at, :timestamp
```
pixeltrix: Not sure about the name - `government_response_at` or `official_response_at` are alternatives. Ask @h-lame what he thinks.

features/step_definitions/petition_steps.rb
```diff
@@ -43,7 +43,18 @@
...
+    title: title,
+    closed_at: state == 'closed' ? 1.day.ago : 6.months.from_now,
+    response: 'Government Response',
+    gov_response_at: govt_response_days_ago.to_i.days.ago
+  }
+  FactoryGirl.create(:open_petition, petition_attributes)
+end
+
 Given /^I have created an e-petition$/ do
+
```
pixeltrix: No need for the whitespace here

pixeltrix:
:+1:

db/migrate/20150612104023_add_gov_response_at_to_petitions.rb
h-lame: It might need to be ``parliament_response_at`` because it's not just the government that would be involved, right?  @yahoopete ?

otlaitil: Yeah, I just copied the name from the story ACs. `parliament_response_at` sounds good to me, I will change it. :+1:

app/models/petition.rb
otlaitil: :+1: for noticing this.

otlaitil:
@pixeltrix @h-lame thank you for the comments, they've been addressed.

app/models/petition.rb
```diff
@@ -28,6 +28,7 @@ class Petition < ActiveRecord::Base
 
   has_perishable_token called: 'sponsor_token'
 
+  before_save :stamp_parliament_response_at, if: -> { response.present? && response_was.nil? }
   after_create :set_petition_on_creator_signature
```
davidbasalla: hey just curious, what is this response_was.nil? function? Where is defined?

mybits:
:+1:

app/models/petition.rb
otlaitil: @davidbasalla `response_was` (or "#{attr}_was") is an `ActiveRecord` feature, which is useful when you want to check if the value has been changed. All model attributes have these methods. Some examples: https://github.com/rails/rails/blob/3e36db4406beea32772b1db1e9a16cc1e8aea14c/activemodel/lib/active_model/dirty.rb

h-lame: ActiveRecord has a feature called "dirty tracking" where it knows what attributes were changed during the last save.  See http://api.rubyonrails.org/classes/ActiveModel/Dirty.html for more details at the API level and [this railscast](http://railscasts.com/episodes/109-tracking-attribute-changes) (a bit old but covers the basics of the API)

h-lame: Oh, sorry for the duplication here. I wrote my comment before @otlaitil responded - just forgot to hit "go".

davidbasalla: brilliant, thanks!

h-lame:
I've just merged #139 which may impact this PR.  @otlaitil at minimum you'll need a rebase, you might need to refresh somewhat too, sorry!

otlaitil:
@h-lame rebased, please merge it when green?

h-lame:
@otlaitil - Now that we have two fields to represent the response:  ``response`` and ``response_summary``, things like ``Petition.with_response`` and the conditional on ``stamp_parliament_response_at`` need to take both fields into account.

otlaitil:
@h-lame ah that's true, I missed the point from your last comment, sorry. Will fix.

PR #137, opened 2015-06-12
Convert AppConfig and SystemSetting to Site model

pixeltrix:
Since we only have at most a handful of settings it makes more sense to store them in explicit columns than cater for an infinite set. This allows us to cache the model in a thread local for performance and then reload in the application controller to allow for admin updates, e.g:

``` ruby
class ApplicationController < ActionController::Base
  before_filter :reload_site

  protected

  def reload_site
    Site.reload
  end
end
```

This will also be useful for when we implement 'Russian Doll' caching - by incorporating the updated_at column in the cache key we can create a global reset that we can both hook up to an admin function or call programmatically in a deploy script if we need to blow away any existing cache:

``` ruby
class Admin::SettingsController < AdminController
  def reset_cache
    Site.touch(:updated_at)
  end
end
```

Some notes on implementation:

1.  The `before_remove_const` method is a hook that Rails calls when unloading classes before reloading them in development mode. This allows us to clear the cached instance in the thread local, otherwise we would end up with an instance that has a stale class definition.

2.  By using `first_or_create` to load the instance we can automatically seed the database with the correct values by defining the appropriate defaults on the respective columns. Changing to this exposed some failures in our cucumber test setup - it was defaulting to a threshold level of 0 so I had to correct for that.

3.  I've added explicit settings for the debate and response threshold to make clear what each one is for.


h-lame:
Reviewing this makes me think, why don't we just put these 2 numbers into application.yml? Or, why don't we take the threshold figures that are in application.yml out and put them in here?

pixeltrix:
@h-lame the stuff in application.yml is things that don't need to change live but there's probably no reason it couldn't all live in `Site`.

h-lame:
Yeah, looking at application.yml most of it is things that shouldn't really change, but I think these:

```
  sponsor_count_min: 5
  sponsor_count_max: 20
  sponsor_moderation_threshold: 5
  petition_duration: 6
```

are of a similar "shape" to the two attributes on ``Site`` now.

pixeltrix:
Don't merge this yet - it's morphed into a "get this thing working on AWS" so I've got some more commits to push up.

h-lame:
This is really nice.  Gives us loads of site control stuff for "free".

pixeltrix:
Okay, this can be merged now as the AWS environments appear to be working

PR #138, opened 2015-06-12
New create petition steps sequence

mybits:
New petition create steps sequence:
create petition --> check your petition --> sign your petition --> make sure this is right (email replay)

PT: https://www.pivotaltracker.com/n/projects/307301/stories/96866290

### DIFF
--- a/app/models/staged/petition_creator/replay_petition.rb
+++ b/app/models/staged/petition_creator/replay_petition.rb
@@ -1,10 +1,6 @@
 module Staged
   module PetitionCreator
     class ReplayPetition < Staged::Base::Petition
-      include Staged::PetitionCreator::HasCreatorSignature
-
-      class CreatorSignature < Staged::Base::CreatorSignature
-      end
     end
   end
 end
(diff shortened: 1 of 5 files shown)


app/models/staged/petition_creator/replay_petition.rb
```diff
@@ -1,7 +1,6 @@
 module Staged
   module PetitionCreator
     class ReplayPetition < Staged::Base::Petition
-      include Staged::PetitionCreator::HasCreatorSignature
 
       class CreatorSignature < Staged::Base::CreatorSignature
```
h-lame: If we don't include the ``Staged::PetitionCreator::HasCreatorSignature`` anymore, we also don't need the ``Staged::PetitionCreator::ReplayPetition::CreatorSignature``.

h-lame:
One more thing that can be deleted, but :+1: on the rest.

app/models/staged/petition_creator/replay_petition.rb
mybits: Oh yes, absolutely! Will remove

mybits:
thanks @h-lame! redundant class removed here: https://github.com/alphagov/e-petitions/commit/aa2b741c9802e6a4490d63eb861b8f9d82623322

otlaitil:
:+1:

PR #144, opened 2015-06-15
Cleaning rejected petition view by removing redundant info and also add new styling

mybits:
- removed: the signature count, confirmation email section, references to 10k and 100k sections,
created by section

changed: 
* Change the date from deadline to be the date created, and call it Date submitted.
* add the styled 'flash' message
* add a reason for rejection header

PT: https://www.pivotaltracker.com/n/projects/307301/stories/96936096


app/views/petitions/_rejected_petition.html.erb
```diff
@@ -0,0 +1,21 @@
+<h1>
+  <span class="heading-secondary">Rejected petition</span>
+  <%= @petition.title %>
```
alexander-bobin: Is it better to have `petition` is a local variable explicitly passed through or just rely on it being available as `@petition`? I'm not sure

app/views/petitions/show.html.erb
```diff
@@ -12,69 +12,62 @@
...
-<% end if @petition.closed? %>
+<% if @petition.rejected? %>
+  <%= render 'rejected_petition' %>
+<% else %>
 
-<h1>
-  <span class="heading-secondary">Petition</span>
-  <%= @petition.title %>
-</h1>
+  <%= content_tag :p, :class => 'flash-notice' do -%>
```
alexander-bobin: Is it better to have this as another partial too? Maybe someone else has some reckons

app/views/petitions/_rejected_petition.html.erb
h-lame: Providing a partial local param gets my vote.  Relying on globals is a bad idea.

otlaitil:
:+1:

app/views/petitions/_rejected_petition.html.erb
mybits: Thanks @alexander-bobin and @h-lame! Couldn't agree more, fixed here:  https://github.com/alphagov/e-petitions/commit/943535ae3adbc57ce531958fcac954c733c70f60

app/views/petitions/show.html.erb
mybits: You mean the heading? If so, I'm not sure if that wouldn't make the code less readable actually..

alexander-bobin: Nope. I meant that we have a partial which represents a rejected petition. Maybe we should have a partial that represents open or not-rejected. It's more of a question to see what others think.

h-lame: It might make it clearer if all the stuff in the ``else`` branch was in a partial, yeah.

mybits:
just been :recycle:

app/views/petitions/show.html.erb
mybits: Ok, will do! Any suggestion how to name this partial? I could only come with names like `view_petition`,  `petition_show`, `not_rejected_petition` - and I don't think they're good enough :thought_balloon:

alexander-bobin: Ah naming. I'd either call it `_petition.html.erb` or `_show_petition.html.erb` and use `_show_rejected_petition.html.erb`

mybits: Like the `petition_show` and `rejected_petition_show` idea :) Will implement those.

mybits:
@alexander-bobin, @h-lame and @otlaitil thanks for your comments! I've added the additional partial in the last commit.

alexander-bobin:
LGTM :shipit:

davidbasalla:
looks good! :+1:

PR #146, opened 2015-06-16
Fix view rejected petitions in admin

davidbasalla:
for https://www.pivotaltracker.com/story/show/97006760

The issue was that rejected petitions did not show up when searching for them. This appears have been caused by use of 'moderated' petition scope which only returns petitions with states OPEN and HIDDEN.
Wondering whether it was intentional to not have REJECTED as part of this. But there are no tests to confirm this, so just changed it to use a new scope for 'selectable' states (OPEN, CLOSED, REJECTED, HIDDEN) which fixes the bug. Have also added more comprehensive testing for these state searches.

### DIFF
--- a/app/controllers/admin/petitions_controller.rb
+++ b/app/controllers/admin/petitions_controller.rb
@@ -2,7 +2,7 @@ class Admin::PetitionsController < Admin::AdminController
   respond_to :html
 
   def index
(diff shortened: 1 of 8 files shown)


davidbasalla:
Thanks for the comments, have fixed and rebased

PR #148, opened 2015-06-16
Add response date to with-response listview UI

otlaitil:
Fix for: https://www.pivotaltracker.com/story/show/96278298

Also refactors petition result items to render a separate partial depending on the facet used in the search. I'm not sure if this is a good idea or not.

### DIFF
--- a/app/assets/stylesheets/petitions/views/shared/_petition-item.scss
+++ b/app/assets/stylesheets/petitions/views/shared/_petition-item.scss
@@ -16,6 +16,9 @@
   .panel-indent {
     margin: 0;
   }
+  p {
+    margin-bottom: 0;
+  }
 }
 .petition-item-existing {
   a {
--- a/app/views/search/petitions/shared/_signature_count.html.erb
+++ b/app/views/search/petitions/shared/_signature_count.html.erb
@@ -1,3 +1,3 @@
 <p>
-  <%= "#{number_with_delimiter(petition.signature_count)} signatures"  %>
+  <%= signature_count(:default, petition.signature_count) %>
 </p>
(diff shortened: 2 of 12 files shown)


h-lame:
I'm with you @otlaitil, I'm on the fence about the partials (mostly because they're not that different).  Still, not on the fence enough to say we should undo it.  Get's a :+1: from me.

mybits:
Yeah, partials party! LGTM :+1:

app/assets/stylesheets/petitions/views/shared/_petition-item.scss
```diff
@@ -16,6 +16,9 @@
   .panel-indent {
     margin: 0;
   }
+  p {
+    margin-bottom: 0;
```
alexander-bobin: This probably effects a few different `.petition-item`s. Should be OK cause we have margin-bottom on the item itself, but worth cross checking.

app/views/search/petitions/shared/_signature_count.html.erb
```diff
@@ -0,0 +1,3 @@
+<p>
+  <%= "#{number_with_delimiter(petition.signature_count)} signatures"  %>
```
alexander-bobin: We should probably be using the `signature_count` helper here

otlaitil: :+1:

app/assets/stylesheets/petitions/views/shared/_petition-item.scss
otlaitil: Seems to look OK in other places as well.

otlaitil:
Rebased & PR comments addressed.

alexander-bobin:
LGTM

PR #151, opened 2015-06-17
Change epetition wording to petition

davidbasalla:
For https://www.pivotaltracker.com/story/show/97092986

Other than lots of renaming, the main change is the use of locale definitions for email signoff, making it easier to change it in future.

### DIFF
--- a/app/controllers/signatures_controller.rb
+++ b/app/controllers/signatures_controller.rb
@@ -32,7 +32,7 @@ class SignaturesController < ApplicationController
     @petition = @signature.petition
 
     if @signature.validated?
-      flash[:notice] = "Thank you. Your signature has already been added to the <span class='nowrap'>e-petition</span>."
+      flash[:notice] = "Thank you. Your signature has already been added to the petition."
       redirect_to signed_petition_signature_url(@petition, @signature.perishable_token) and return
     end
 
--- a/app/views/archived/petitions/index.html.erb
+++ b/app/views/archived/petitions/index.html.erb
@@ -1,16 +1,16 @@
 <div class="cta_block" id="search_pet_cta_block">
-  <h1>Archived e-petitions</h1>
+  <h1>Archived petitions</h1>
 
   <form method="get" action="/archived/petitions/search">
-    <label for="search">Search for an archived e-petition</label>
+    <label for="search">Search for an archived petition</label>
     <input type='text' name="q" id="search" />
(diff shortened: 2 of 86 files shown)


app/controllers/signatures_controller.rb
```diff
@@ -32,7 +32,7 @@ def verify
     @petition = @signature.petition
 
     if @signature.validated?
-      flash[:notice] = "Thank you. Your signature has already been added to the <span class='nowrap'>e-petition</span>."
+      flash[:notice] = "Thank you. Your signature has already been added to the <span class='nowrap'>petition</span>."
```
yahoopete: We don't need the class any more as the word isn't hyphenated.... So it won't wrap.

app/views/archived/petitions/index.html.erb
```diff
@@ -1,16 +1,16 @@
...
-  <%= link_to 'How e-petitions work', help_path, :class => 'large_link' %>
-  <%= link_to "Create a new e-petition", check_petitions_path, :class => 'link_button large_link_button' %>
+  <h2>What are petitions?</h2>
+  <p>Petitions <span class="nowrap">are</span> an easy, personal way for you to influence government and Parliament in the UK. You can create a petition about anything that the government is responsible for and if it gets at least 100,000 signatures, it will be considered for debate in the House of Commons.  You can find more information about how the House of Commons deals with petitions on the <%= new_window_link_to "Backbench Business Committee website", "http://www.parliament.uk/bbcom" %>.</p>
```
yahoopete: The span was already in the wrong place....  But actually,  rather than fix,  we don't need this block for archived so we should just remove it all.

app/views/petitions/create/_your_details_ui.html.erb
```diff
@@ -6,7 +6,7 @@
 <%= render 'petitions/create/error_summary', f: f %>
 
 <%= f.fields_for :creator_signature, petition.creator_signature do |csf| %>
-  <%= render :partial => 'signatures/form', :locals => { :f => csf, :signature => csf.object, :context => 'create-petition' } %>
+  <%= render :partial => 'signatures/form', :locals => { :f => csf, :signature => csf.object, :context => 'creatpetition' } %>
```
yahoopete: Overly aggressive find and replace?

davidbasalla: doh!

h-lame:
Thorough work!

PR #153, opened 2015-06-17
Update README: data generation & sql structure

otlaitil:


h-lame:
Good catch!  Always good to keep the readme up to date.

PR #145, opened 2015-06-16
Debate outcomes

h-lame:
For: https://www.pivotaltracker.com/story/show/96654388

Add a debate outcome model attached (has_one) to petition.  Allow admins to fill in the details of this model via a new form.  Publish that information - if available - on the petition show in the right place (under the 100,000 signatures info block).

PR #154, opened 2015-06-18
Moderator can edit petition before publishing

mybits:
- Style is not implemented, but a template follows the markup we're gonna use while adding new style to the admin.
- After updating petition moderator is redirected to back to the petition page (not the index page)

PT: https://www.pivotaltracker.com/n/projects/307301/stories/96865488

### DIFF
--- a/app/controllers/admin/petition_details_controller.rb
+++ b/app/controllers/admin/petition_details_controller.rb
@@ -1,0 +1,20 @@
+class Admin::PetitionDetailsController < Admin::AdminController
+  respond_to :html
+  before_action :find_petition
+
+  def show
+    if @petition.in_todo_list?
+    else
+      raise ActiveRecord::RecordNotFound
+    end
+  end
+
+  def update
+  end
+
+  private
+
+  def find_petition
+    @petition = Petition.find(params[:petition_id])
+  end
+end
--- a/app/controllers/admin/petition_details_controller.rb
+++ b/app/controllers/admin/petition_details_controller.rb
@@ -1,20 +1,27 @@
 class Admin::PetitionDetailsController < Admin::AdminController
   respond_to :html
-  before_action :find_petition
+  before_action :fetch_petition
 
   def show
-    if @petition.in_todo_list?
-    else
-      raise ActiveRecord::RecordNotFound
-    end
   end
 
   def update
+    if @petition.update_attributes(petition_params)
+      flash[:notice] = 'Petition has been successfully updated'
+      redirect_to edit_admin_petition_path(@petition)
+    else
+      render :show
+    end
   end
 
   private
 
-  def find_petition
-    @petition = Petition.find(params[:petition_id])
+  def fetch_petition
+    @petition = Petition.todo_list.find(params[:petition_id])
+    #raise ActiveRecord::RecordNotFound unless @petition.in_todo_list?
+  end
+
+  def petition_params
+    params.require(:petition).permit(:title, :action, :description)
   end
 end
--- a/app/models/petition.rb
+++ b/app/models/petition.rb
@@ -198,6 +198,10 @@ class Petition < ActiveRecord::Base
     self.open? || self.closed?
   end
 
+  def in_todo_list?
+    TODO_LIST_STATES.include? state
+  end
+
   def state_label
     if (self.closed?)
       CLOSED_STATE
--- a/app/models/petition.rb
+++ b/app/models/petition.rb
@@ -170,10 +170,6 @@ class Petition < ActiveRecord::Base
     self.state == SPONSORED_STATE
   end
 
-  def collecting_sponsors?
-    self.state == VALIDATED_STATE || self.state == PENDING_STATE
-  end
-
   def open?
     self.state == OPEN_STATE
   end
--- a/app/views/admin/petitions/edit.html.erb
+++ b/app/views/admin/petitions/edit.html.erb
@@ -2,7 +2,11 @@
 
   <%= render :partial => 'petition_details', :locals => { :f => f } %>
 
-  <div>
+  <div class="row">
+    <%= link_to 'Edit petition content', admin_petition_petition_details_path(@petition) %>
+  </div>
+
+  <div class="row">
     <%= submit_tag 'Publish this petition', :data => { :confirm => "Are you sure you want to publish this petition?" } %>
   </div>
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -65,6 +65,7 @@ Rails.application.routes.draw do
         patch :take_down
       end
       resource 'debate-outcome', only: [:show, :update], as: :debate_outcome, controller: :debate_outcomes
+      resource :petition_details, :only => [:show, :update]
     end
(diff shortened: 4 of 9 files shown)


app/controllers/admin/petition_details_controller.rb
```diff
@@ -0,0 +1,30 @@
+class Admin::PetitionDetailsController < Admin::AdminController
+  respond_to :html
+  before_action :find_petition
+
+  def show
+    if @petition.in_todo_list?
```
otlaitil: We could use a guard instead of an empty `if` statement, so the whole action would end up reading like this:

`raise ActiveRecord::RecordNotFound unless @petition.in_todo_list?`

I think you could extract line this to a separate method and use it in `update` action as well. Currently `update` will accept editing a petition that is not in todo-list, if you tamper the request (probably not a big deal since this is an admin feature).

app/models/petition.rb
```diff
@@ -193,6 +189,10 @@ def closed?
     self.state == OPEN_STATE && self.closed_at <= Time.current
   end
 
+  def in_todo_list?
+    state == SPONSORED_STATE || state == VALIDATED_STATE || state == PENDING_STATE
```
otlaitil: Nitpick: we could simplify this to: `[SPONSORED_STATE, VALIDATED_STATE, PENDING_STATE].include? state`

config/routes.rb
```diff
@@ -64,6 +64,7 @@
         patch :update_response
         patch :take_down
       end
+      resource :petition_details, :only => [:show, :update], as: :petition_details, controller: :petition_details
```
otlaitil: I think you could drop these: `, as: :petition_details, controller: :petition_details` from the resource definition

otlaitil:
LGTM! :+1: One thing I am not completely sure about is having a separate restful controller for just updating the details... because details is not really a separate resource. But OK for me!

app/controllers/admin/petition_details_controller.rb
mybits: Good idea! Will do!

app/models/petition.rb
mybits: oh yes, much better :+1:

config/routes.rb
mybits: I guess I can :)

app/controllers/admin/petition_details_controller.rb
h-lame: Could also just change the ``find_petition`` method to use ``Petition.todo_list.find(params[:petition_id])``

app/views/admin/petitions/edit.html.erb
```diff
@@ -2,7 +2,11 @@
 
   <%= render :partial => 'petition_details', :locals => { :f => f } %>
 
-  <div>
+  <div class="row">
+    <%= link_to 'Edit petition content', admin_petition_petition_details_path(@petition) %>
+  </div>
```
h-lame: Now that #145 is merged we can put this action on the show page with the other actions.

mybits:
Thanks @otlaitil for your comments! All addressed here: https://github.com/alphagov/e-petitions/commit/ec001001c79f8fd9a28b85f83ed0c0c659318b6a and here: https://github.com/alphagov/e-petitions/commit/eaec48bba3ea51aac830cf469a4e0d1d8119fd29
The reason for having a separate controller was to not put more actions into `petitions_controller` so it doesn't grow more, especially that the edit/update actions are used for publishing/rejecting petitions.
I think there might be a plan to refactor  `petition_controller` while working on admin stories in general.

features/step_definitions/moderation_steps.rb
```diff
@@ -3,6 +3,11 @@
   visit edit_admin_petition_path(@petition)
 end
 
+When /^I click at the sponsored petition with title: "([^"]*)", that has action: "([^"]*)" and description: "([^"]*)"$/ do |title, action, description|
+  @sponsored_petition = FactoryGirl.create(:sponsored_petition, title: title, action: action, description: description)
+  visit edit_admin_petition_path(@sponsored_petition)
+end
```
h-lame: This step merges data creation and user action.  It's not a major sin or anything, but we should be wary of it.

However, I think my main concern is the name "I click at..." which suggests a user looking at something on screen and clicking on it, when in fact we just point their browser directly at the edit page for them.  Maybe just change the title to "I visit ...." ?

app/views/admin/petitions/edit.html.erb
mybits: Great! Will do!

spec/controllers/admin/petition_details_controller_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+
+          it 'updates the petition' do
+            do_update
+            petition.reload
+            expect(petition).to be_present
+            expect(petition.title).to eq('New title')
+            expect(petition.action).to eq('New action')
+            expect(petition.description).to eq('New description')
+          end
+        end
```
h-lame: Be worth adding a spec to show that these 3 are the only attributes you can manipulate via this controller and action.

h-lame:
Good stuff!  Suggested a couple of minor tweaks, but I like it.

features/step_definitions/moderation_steps.rb
mybits: Good point, will change!

spec/controllers/admin/petition_details_controller_spec.rb
mybits: Ha yes! Didn't think about it and that'll be definitely a better test coverage, will add!

app/views/admin/petitions/edit.html.erb
mybits: Oh no, actually I can't do it - as the editing content is only for petitions that are in moderation so it doesn't include petitions that are already open/rejected/etc

h-lame: You can put a guard clause on showing the link though, so it only appears for the petitions you're allowed to edit in this way (it's what I did for the debate outcomes link).

mybits:
@otlaitil and @h-lame thanks for all your comments. All have been addressed so that is ready for a final look :)

app/views/admin/petitions/edit.html.erb
h-lame: Turns out this action is kinda special and so won't appear on the "main" action list anyway.  No need to move it.

app/controllers/admin/petition_details_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+    else
+      render :show
+    end
+  end
+
+  private
+
+  def fetch_petition
+    @petition = Petition.find(params[:petition_id])
+    raise ActiveRecord::RecordNotFound unless @petition.in_todo_list?
```
h-lame: Minor thing, but if we do:

```ruby
def fetch_petition
  @petition = Petition.todo_list.find(params[:petition_id])
end
```

We don't need the guard clause at the end as AR will raise the exception for us by going through the scope.

PR #156, opened 2015-06-18
Send emails to signatories about debate outcomes

h-lame:
For: https://www.pivotaltracker.com/story/show/96655040

Mostly the work here is around refactoring the existing ``EmailThresholdResponseJob`` to make the email behaviour more extensible.  See 33c3326b6f85935cc5dfd9ea71e81619bb5ccfbf for a description of that behaviour.  We add an ``EmailRequestedReceipt`` object which is attached to ``Petition`` and this holds the timestamp for when each email type was requested.  There is a corresponding ``EmailSentReceipt`` object attached to each ``Signature`` that contains the timestamp for when each email type was last sent to the signature.  Currently the receipt objects have two timestamps ``government_response`` for the ``PetitionMailer.notify_signer_of_threshold_response`` email and ``debate_outcome`` for the ``PetitionMailer.notify_signer_of_debate_outcome`` email.

### DIFF
--- a/app/jobs/email_debate_outcomes_job.rb
+++ b/app/jobs/email_debate_outcomes_job.rb
@@ -1,19 +1,12 @@
-class EmailDebateOutcomesJob < ActiveJob::Base
+class EmailDebateOutcomesJob < EmailPetitionSignatories::Job
   def self.run_later_tonight(petition)
     petition.set_email_requested_at_for('debate_outcome', to: Time.current)
-    EmailPetitionSignatories.run_later_tonight(self, petition, petition.get_email_requested_at_for('debate_outcome'))
+    super(petition, petition.get_email_requested_at_for('debate_outcome'))
   end
-
-  queue_as :default
 
   def perform(petition, requested_at_string, mailer = PetitionMailer.name, threshold_logger = nil)
-    @petition = petition
     @mailer = mailer.constantize
-    worker(requested_at_string, threshold_logger).do_work!
-  end
-
-  def worker(requested_at_string, threshold_logger)
-    EmailPetitionSignatories::Worker.new(self, @petition, requested_at_string, threshold_logger)
+    worker(petition, requested_at_string, threshold_logger).do_work!
   end
 
   def timestamp_name
--- a/app/jobs/email_debate_outcomes_job.rb
+++ b/app/jobs/email_debate_outcomes_job.rb
@@ -4,9 +4,9 @@ class EmailDebateOutcomesJob < EmailPetitionSignatories::Job
     super(petition, petition.get_email_requested_at_for('debate_outcome'))
   end
 
-  def perform(petition, requested_at_string, mailer = PetitionMailer.name, threshold_logger = nil)
+  def perform(petition, requested_at_string, mailer = PetitionMailer.name, logger = nil)
     @mailer = mailer.constantize
-    worker(petition, requested_at_string, threshold_logger).do_work!
+    worker(petition, requested_at_string, logger).do_work!
   end
 
   def timestamp_name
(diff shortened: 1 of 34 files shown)


app/jobs/email_petition_signatories.rb
```diff
@@ -0,0 +1,85 @@
+class EmailPetitionSignatories
+  class PleaseRetry < StandardError; end
+
+  def self.run_later_tonight(job_class, petition, requested_at, *extra_args)
```
otlaitil: Maybe it's just me, but it too quite a long time to understand what's going on here between the jobs, `EmailPetitionSignatories` and the `Worker`.

Somehow I feel that `self.run_later_tonight` and `self.later_tonight` should not be the responsibility of this class.

What do you think if this class would act just as the worker - and those two methods could be extracted to a separate class, something like `EmailPetitionSignatoriesJob`, which would be a superclass for `EmailDebateOutcomesJob` and `EmailDebateOutcomesJob`? The jobs then would no longer have to communicate with this class to queue themselves.

h-lame: Yeah, you're right, those methods don't really belong here - having to supply the job class should've been a hint to me!  A previous version that I moved away from actually had ``EmailPetitionSignatories::Worker`` be the ``ActiveJob`` instance and the two current ``ActiveJob`` instances (``EmailDebateOutcomesJob`` and ``EmailThresholdResponseJob``) weren't actually ``ActiveJob`` subclasses at all, they just configured the ``EmailPetitionSignatories::Worker`` by passing in loads of arguments.  I didn't really like that (especially that something called ``BlahJob`` wasn't actually a Job) so moved away, but didn't refactor these two methods.  Will do!

h-lame: Responded in 58aa2db86096d49078b93a46293ff719e2e56652 (commit message has some reasoning in it so you might want to read that rather than review the total diff).

db/migrate/20150617114935_create_email_requested_receipts.rb
```diff
@@ -0,0 +1,32 @@
+class CreateEmailRequestedReceipts < ActiveRecord::Migration
+  def change
+    create_table :email_requested_receipts do |t|
+      t.references :petition, index: true, foreign_key: true
```
otlaitil: Just out of interest, I've been thinking to try PostgreSQL's `jsonb` datatype with Rails (seems that the support is pretty good with v. 4.2) some day - what do you think, could it suit this use case? Seems that the receipt implementation is very dynamic apart from the db side.

otlaitil:
Looks good! :+1:

app/jobs/email_debate_outcomes_job.rb
```diff
@@ -0,0 +1,19 @@
+class EmailDebateOutcomesJob < EmailPetitionSignatories::Job
+  def self.run_later_tonight(petition)
+    petition.set_email_requested_at_for('debate_outcome', to: Time.current)
+    super(petition, petition.get_email_requested_at_for('debate_outcome'))
+  end
+
+  def perform(petition, requested_at_string, mailer = PetitionMailer.name, threshold_logger = nil)
```
otlaitil: This logger could be renamed.

h-lame: Probably just to logger, yeah?  I'll do that.

h-lame:
Rebased and responded to comments.  There's now another story working on top of this branch so I guess we should merge soon?

db/migrate/20150617114935_create_email_requested_receipts.rb
h-lame: It hadn't even crossed my mind - I keep forgetting we're using a fun db.  Maybe something to refactor towards if we add lots more emails?

otlaitil: Yup, it could be worth trying!

PR #159, opened 2015-06-18
Changed 'parliament' response to 'Government' response

yahoopete:
Also changed the admin petition show page so that the links to the actions aren't "Add a xxxxxxxx" as that isn't useful after the response is added.


h-lame:
Squashed this a bit, sorry to reduce your contribution-by-commit count Pete!  Will merge when :green_heart:

yahoopete:
Damn it... you're on to me.

PR #160, opened 2015-06-18
Optional supporting details and rename Petition DB fields to match UI

h-lame:
For: https://www.pivotaltracker.com/story/show/97178596 and https://www.pivotaltracker.com/story/show/96647416 (yes I know it's bad form to do 2 stories in 1 PR sorry!)

This does touch many files so chances are this will either need hard rebasing as more things are merged in, or require other things to do lots of rebasing if they are merged in afterwards so we need to take some care.


h-lame:
Oh, look at me not being consistent with my commit titles.  Oops.

otlaitil:
:+1:

davidbasalla:
Looks good! Out of curiosity, how did you go about doing the bulk renaming of fields?

h-lame:
@davidbasalla nothing fancy: just find and replace in my editor + running the tests to see what I'd missed (searching for ``attribute:``, ``attribute =>``, ``:attribute``, ``.attribute`` rather than just ``attribute`` was useful for title as that is used in lots of places to mean other things).

mybits:
:+1: :cake:

h-lame:
I grow weary of rebasing this, can we merge if it goes green?

PR #162, opened 2015-06-19
Removing scheduled debate link from edit response view

yahoopete:
We don't need this on this page right now, it may come back (where we can go from any edit page to any other) but as it's the only one right now, it feels a bit odd.


alanth:
:+1:

PR #164, opened 2015-06-19
Create a new list view for 'Petitions awaiting a Government response'

pixeltrix:
When petitions have a signature count above a threshold (currently 10,000) they are meant to have a response from the Government. This adds a list for viewing those petitions which have reached this threshold but have yet to receive a response.

https://www.pivotaltracker.com/story/show/96278100

As part of this I needed to make signature counts live.

https://www.pivotaltracker.com/story/show/96434874


h-lame:
Looks good.  Merging when green.

PR #165, opened 2015-06-19
Tell the signatory their actual number

pixeltrix:
Rather than a generic thank you message, display the actual number at which the person signed by counting the number of signatures with a earlier validated_at date.

https://www.pivotaltracker.com/story/show/96865510


pixeltrix:
Review as okay by @h-lame

PR #166, opened 2015-06-22
Add foreign keys to existing tables

pixeltrix:
Rails 4.2 added built-in support for foreign keys and referential integrity so we should take advantage of this to ensure that our database remains consistent even under heavy load.

### DIFF
--- a/app/models/debate_outcome.rb
+++ b/app/models/debate_outcome.rb
@@ -2,6 +2,5 @@ class DebateOutcome < ActiveRecord::Base
   belongs_to :petition, touch: true
 
   validates :petition, :debated_on, presence: true
-  validates :petition_id, uniqueness: true
   validates :transcript_url, :video_url, length: { maximum: 500 }
 end
--- a/spec/models/signature_spec.rb
+++ b/spec/models/signature_spec.rb
@@ -79,84 +79,81 @@ RSpec.describe Signature, type: :model do
     end
 
     describe "uniqueness of email" do
-      before do
-        FactoryGirl.create(:signature, :name => "Suzy Signer",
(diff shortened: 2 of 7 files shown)


spec/models/signature_spec.rb
```diff
@@ -79,84 +79,81 @@
...
+          expect(signature).to have_valid(:email)
+        end
+
+        it "ignores extra whitespace at the end of the name" do
+          signature = FactoryGirl.build(:signature, attributes.merge(name: "Suzy Signer "))
+          expect(signature).not_to have_valid(:email)
+        end
+
+        it "ignores extra whitespace at the beginning of the name" do
+          signature = FactoryGirl.build(:signature, attributes.merge(name: "Suzy Signer "))
```
pixeltrix: This is meant to be at the beginning - ugh :disappointed:

app/models/debate_outcome.rb
```diff
@@ -2,6 +2,5 @@ class DebateOutcome < ActiveRecord::Base
   belongs_to :petition, touch: true
 
   validates :petition, :debated_on, presence: true
-  validates :petition_id, uniqueness: true
```
h-lame: There's a uniqueness constraint on ``debate_outcomes.petition_id`` so (race conditions with ``validates_uniqueness_of`` notwithstanding) isn't it still worth having this validation?

otlaitil:
Nice! :+1:

h-lame:
Good spot.  Most of these are my fault too as I did the PRs with that added most of these tables.  Will endeavour to pay more attention on the next models I add.  Don't know how I managed to get these wrong, but ``EmailRequestedReceipt`` and ``EmailSentReceipt`` correct.

mybits:
:+1:

PR #167, opened 2015-06-22
Add a page title locale key for awaiting response

pixeltrix:
When the list of petitions awaiting a reponse was added in e5a2d1a it was missing a key in the config/locales/petitions.en-GB.yml file for the page title - this commit adds that missing key.


otlaitil:
:+1:

davidbasalla:
cool

PR #168, opened 2015-06-22
Moderation on the show page in admin

h-lame:
For: https://www.pivotaltracker.com/story/show/97012430

Not styled beyond what we do by default so I might ask for some advice from @alexander-bobin on it.  There may be more commits, but that shouldn't stop you from reviewing now.


otlaitil:
LGTM :+1:

mybits:
LGTM :dart:

app/views/admin/moderation/_form.html.erb
```diff
@@ -0,0 +1,17 @@
+<%= form_for petition, url: admin_petition_moderation_path(petition), method: :patch do |f| -%>
+  <%= form_row do %>
+    <%= label_tag :moderation_approve, nil do %>
```
alexander-bobin: GDS tend to wrap radio and checkbox groups in a `<fieldset class="inline">` but that can be done later.

alexander-bobin:
The markup looks fine to me. It won't be a big jump to work it into shape when the new styles come into play

PR #170, opened 2015-06-22
Add .inline-search, apply styles

henryhadlow:
Style the search box and button to appear as one horizontal unit


henryhadlow:
Replaced with https://github.com/alphagov/e-petitions/pull/173

PR #171, opened 2015-06-22
When showing a petition don't show no "more details" when there are none

h-lame:
We also standardise how we were showing the additional_details to always:

1. use the ``<details>/<summary>`` tags pattern
2. push the content through simple_format and autolink

For bug fix on: https://www.pivotaltracker.com/story/show/97178596


otlaitil:
LGTM :+1:

PR #176, opened 2015-06-22
Petitions awaiting debate list

mybits:
PT: https://www.pivotaltracker.com/n/projects/307301/stories/96278800

### DIFF
--- a/app/models/petition.rb
+++ b/app/models/petition.rb
@@ -42,6 +42,8 @@ class Petition < ActiveRecord::Base
   facet :rejected, -> { for_state(REJECTED_STATE).reorder(created_at: :desc) }
   facet :all, -> { reorder(signature_count: :desc) }
 
+  facet :awaiting_debate_date, -> { where(state: OPEN_STATE).awaiting_debate_date.without_debate_outcome.reorder(debate_threshold_reached_at: :asc) }
+
   # = Relationships =
   belongs_to :creator_signature, :class_name => 'Signature'
   accepts_nested_attributes_for :creator_signature
@@ -100,6 +102,10 @@ class Petition < ActiveRecord::Base
   scope :with_response, -> { where.not(response_summary: nil, response: nil) }
   scope :with_debate_outcome, -> { joins(:debate_outcome) }
 
+  scope :without_debate_outcome, -> { where.not(id: DebateOutcome.select(:petition_id).uniq) }
+
+  scope :awaiting_debate_date, ->  { where.not(debate_threshold_reached_at: nil) }
+
   def self.awaiting_response
     where(state: OPEN_STATE, response: nil, response_summary: nil).where.not(response_threshold_reached_at: nil)
   end
@@ -134,6 +140,10 @@ class Petition < ActiveRecord::Base
       updates << "response_threshold_reached_at = :now"
     end
 
+    if at_threshold_for_debate?
+      updates << "debate_threshold_reached_at = :now"
+    end
+
     if update_all([updates.join(", "), now: time]) > 0
       self.reload
     end
@@ -142,6 +152,12 @@ class Petition < ActiveRecord::Base
   def at_threshold_for_response?
     unless response_threshold_reached_at?
       signature_count >= Site.threshold_for_response - 1
+    end
+  end
+
+  def at_threshold_for_debate?
+    unless debate_threshold_reached_at?
+      signature_count >= Site.threshold_for_debate - 1
     end
   end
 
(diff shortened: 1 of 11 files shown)


features/step_definitions/petition_steps.rb
```diff
@@ -65,7 +65,13 @@
   FactoryGirl.create(:open_petition, petition_attributes)
 end
 
-Given(/^a petition "(.*?)" exists and hasn't passed the threshold for a response$/) do |action|
+Given(/^a petition "([^"]*)" exists with a signature count of (\d+) and not set a debate date$/) do |title, count|
+  p = FactoryGirl.create(:open_petition)
+  p.update_attribute(:singature_count, count)
+  p.debate_outcome.debate_on == nil
+end
```
h-lame: I suspect this step isn't used because the attribute is actually ``debated_on`` so it would error.  Also, setting the a field to nil doesn't actually save the object, which is good because if it was the right field, setting debated_on to nil would make it invalid as debate outcomes need to have a debated_on date.  Probably just wanted to do ``p.debate_outcome = nil`` like in the other new steps (because, unlike setting attributes to ``nil``, setting a has_one to nil *does* save to the DB).  As it is I think we can just delete this step.

app/models/petition.rb
```diff
@@ -100,6 +102,10 @@ class Petition < ActiveRecord::Base
   scope :with_response, -> { where.not(response_summary: nil, response: nil) }
   scope :with_debate_outcome, -> { joins(:debate_outcome) }
 
+  scope :without_debate_date, -> { where.not(id: DebateOutcome.select(:petition_id).uniq) }
```
h-lame: Maybe change this to ``without_debate_outcome`` so it's clearer that it's the inverse of that scope?

h-lame: I think we could do this without the ``DebateOutcome.select(:petition_id).uniq`` too.  Something like:

```ruby
eager_load(:debate_outcome).where(debate_outcomes: {id: nil})
```

This does a left outer join on debate_outcomes and throws away the rows that have a joined table.  However, by using eager_load we force AR to load the debate_outcome objects when we know there won't be any, so we might want to look at using ARel (or just a string) to do the join without asking AR to load the objects.

db/structure.sql
```diff
@@ -945,5 +953,7 @@ INSERT INTO schema_migrations (version) VALUES ('20150619075903');
 
 INSERT INTO schema_migrations (version) VALUES ('20150619090833');
 
+INSERT INTO schema_migrations (version) VALUES ('20150622083615');
```
h-lame: Not a big thing, but this probably wants to be at the end of the file as the migrations should be inserted in order.  Anyone running rake db:migrate will probably end up with things in a different order and we'll have dirty db/structure.sql files.  Probably just need to reset your db before you run the migrations.

spec/models/petition_spec.rb
```diff
@@ -334,6 +334,22 @@
...
+
+      it 'returns only petitions that reached the debate threshold' do
+        expect(Petition.awaiting_debate_date).to include(@p2)
+      end
+
+      it 'doesn\'t include petitions that has the debate date' do
+        expect(Petition.awaiting_debate_date).not_to include(@p3)
+      end
+    end
+
```
h-lame: Be nice to have a spec for the instance version of this ``at_threshold_for_debate?`` too.  Just for clarity (and to make sure it's symmetrical with the scope version).

app/models/petition.rb
otlaitil: What's wrong with using `DebateOutcome.select(:petition_id).uniq` here? The scope seems to translate to a good looking SQL with a subquery (and from what I have heard Postgres can optimise it and decide to use a join under the hood if it sees it beneficial?)

h-lame: Oh! I had no idea it would use a subquery.  I thought it would do two queries.  That's really neat, thanks for pointing it out @otlaitil!  Ignore me then, we can leave the implementation as-is.  

Now I have to go work out if this is PG specific or something AR does now whenever it can.

mybits: Ha yeah, it was hard to find so good looking scope, glad it can stay :)

db/structure.sql
mybits: Sure, my db went a bit crazy. Will fix that.

features/step_definitions/petition_steps.rb
mybits: Ah! Great catch, thanks, will fix that!

spec/models/petition_spec.rb
mybits: Absolutely, I'll add it!

mybits:
@h-lame and @otlaitil thanks for your comments, all should be addressed in relevant commits

h-lame:
Rebase needed, but we can merge when rebased+green.

PR #177, opened 2015-06-22
Don't show petition creator on closed petitions

pixeltrix:
To prevent future embarrassment for petition creators, once a petition has closed then no longer show their name. An example of this would be am employer searching on a potential employee's name and getting back a petition they created 2-3 years ago as one of the top results.

https://www.pivotaltracker.com/story/show/97192450


h-lame:
:+1:

otlaitil:
:+1:

mybits:
aand the 3rd :+1:

PR #179, opened 2015-06-22
Fix signature number after signing

pixeltrix:
Commit 6a5849d added a class method that should've been scoped to a petition when calculating the number of signature. This adds that scope and a regression test to ensure that it doesn't reoccur.

https://www.pivotaltracker.com/story/show/96865510


h-lame:
:+1:

mybits:
LGTM :+1:

PR #180, opened 2015-06-23
Accessibility improvements

pixeltrix:
Fix the accessibility concerns with the markup raised by GDS:

- Add `role="main"` to the `main` element
- Add `aria-labelledby` attributes to all `<section>` elements
- Add a `h1` to home page wrapped around Petitions in the header
- Add a `aria-labelledby` attribute to `<nav id="other-search-lists">`
- Add `role="navigation"` attribute to `<nav>` elements
- Remove `aria-hidden` from elements that are empy

There was a suggest to use `aria-describedby` as well but there didn't seem any section where it seemed appropriate.

https://www.pivotaltracker.com/story/show/97282222

### DIFF
--- a/app/views/shared/_header.html.erb
+++ b/app/views/shared/_header.html.erb
@@ -1,8 +1,8 @@
 <header role="banner" id="global-header">
   <div class="wrapper">
     <%= link_to(home_path, :title => 'Go to the petitions homepage', :tabindex => 1) do %>
-      <span class="graphic graphic-portcullis-white" aria-hidden="true"></span> Petitions
+      <span class="graphic graphic-portcullis-white"></span> Petitions
     <% end %>
-	<%= yield :strap_line %>
+    <%= yield :strap_line %>
   </div>
 </header>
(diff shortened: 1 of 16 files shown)


otlaitil:
:+1:

app/views/shared/_header.html.erb
```diff
@@ -1,8 +1,10 @@
 <header role="banner" id="global-header">
   <div class="wrapper">
-    <%= link_to(home_path, :title => 'Go to the petitions homepage', :tabindex => 1) do %>
-      <span class="graphic graphic-portcullis-white" aria-hidden="true"></span> Petitions
-    <% end %>
+    <h1>
```
alexander-bobin: Unfortunately we only want this to be an `h1` when it's the home page. On other pages we have good `h1`s already. This was not fully explained in the story, my bad.

alexander-bobin: @pixeltrix ^^

h-lame:
Rebase pls.

PR #183, opened 2015-06-23
Fix occasionally failing signature spec

otlaitil:
The spec matched the order of the signatures returned by a scope in
signature, altough there was no promise of the order.


h-lame:
Good spot.  Merge this if it's green; although I guess the green might not be triggered by the change - the db order being non-deterministic, but I'm willing to believe it :)

PR #184, opened 2015-06-23
Admin search by keyword

davidbasalla:
Admin search supports search by keywords which is the default choice if search string is not a number or an email). Admin petitions index uses the new search functionality.

### DIFF
--- a/app/controllers/admin/petitions_controller.rb
+++ b/app/controllers/admin/petitions_controller.rb
@@ -2,9 +2,7 @@ class Admin::PetitionsController < Admin::AdminController
   respond_to :html
 
   def index
-    @petitions = Petition.selectable.order(:signature_count)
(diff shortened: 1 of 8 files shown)


otlaitil:
LGTM! :+1:

h-lame:
This does everything it needs to, which is great, but we need to think about the comments on f823f4ef5e1b4c51aced1e43e397b0b8e5f6eb8d before we merge.

davidbasalla:
Have fixed the closed <li> tags issue and rebased. 

After talking to @otlaitil and @pixeltrix, it sounds like the issues that were raised by @h-lame about the hidden facet being available on the front end will be addressed in separate PR's. Should we add another story for that?

h-lame:
Ok, they'll be fixed in another story (see: https://www.pivotaltracker.com/story/show/97724776) so this gets a :+1: for merge from me.

PR #185, opened 2015-06-23
Separate out edit response in admin

h-lame:
For: https://www.pivotaltracker.com/story/show/97006702

The goal here is to break up edit_response and update_response to have separate views and backends.  We addressed the first part of this in #168 and this continues the effort.  Prior to either of these PRs the view had 3 forms on it, the backend dealt with 2 different update types; now each update type is a separate view and backend.

As with #168 I haven't really addressed the styling. We'll see if more needs to be done.


davidbasalla:
Looks good to me :+1:

otlaitil:
LGTM :+1:

PR #186, opened 2015-06-23
Fix bug revalidating sponsor email

davidbasalla:
When revalidating a sponsors email address, it now redirects to the sponsors signed url with an appropriate flash message. Also refactored the signature controller verify method, as it was getting quite complex.

After the refactoring I was wondering if the verification/validation for the sponsor should happen in the sponsor controller rather than the signature controller? Curious to hear other opinions on this!

### DIFF
--- a/app/controllers/signatures_controller.rb
+++ b/app/controllers/signatures_controller.rb
@@ -29,25 +29,11 @@ class SignaturesController < ApplicationController
 
   def verify
     verify_token
-    @petition = @signature.petition
-
-    if @signature.validated?
-      flash[:notice] = "Thank you. Your signature has already been added to the petition."
-      if @signature.sponsor?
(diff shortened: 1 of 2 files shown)


app/controllers/signatures_controller.rb
```diff
@@ -121,5 +111,25 @@ def handle_new_signature(petition)
...
+  end
+
+  def validate_signature
+    if @signature.validated?
+      flash[:notice] = "Thank you. Your signature has already been added to the petition."
+    else
+      @signature.validate!
+    end
+    redirect_to signed_petition_signature_url(@signature.petition, @signature.perishable_token)
+  end
```
h-lame: This is nice; getting rid of the ``and return`` statements that terminate control flow in the middle of the methods makes me happy :)

h-lame:
:+1:

alanth:
:+1: Merging

PR #189, opened 2015-06-24
Bumping the local petition count from 3 to 50

yahoopete:
Pivotal : https://www.pivotaltracker.com/story/show/97656288

A couple of tweaks to the page layout when I was in there.  Questionable use of the locale file by me, so expect a comment here...... :)

### DIFF
--- a/config/locales/petitions.en-GB.yml
+++ b/config/locales/petitions.en-GB.yml
@@ -68,8 +68,8 @@ en-GB:
           other: "<span class=\"count\">%{formatted_count}</span> signatures in the last hour"
       in_your_constituency:
         html:
-          one: "%{formatted_count} signature from your constituency"
-          other: "%{formatted_count} signatures from your constituency"
+          one: "%{formatted_count} signature from "
+          other: "%{formatted_count} signatures from "
       in_total:
         html:
           one: "%{formatted_count} signature total"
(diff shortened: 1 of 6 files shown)


config/locales/petitions.en-GB.yml
```diff
@@ -68,8 +68,8 @@ en-GB:
           other: "<span class=\"count\">%{formatted_count}</span> signatures in the last hour"
       in_your_constituency:
         html:
-          one: "%{formatted_count} signature from your constituency"
-          other: "%{formatted_count} signatures from your constituency"
+          one: "%{formatted_count} signature from "
+          other: "%{formatted_count} signatures from "
```
h-lame: We can specify %{constituency} in the locale:

```yaml
one: "%{formatted_count} signature from %{constituency}"
other: "%{formatted_count} signatures from %{constituency}"
```

Use in the view as:
```ruby
<%= signature_count(:in_your_constituency, petition.constituency_signature_count, constituency: @constituency.name) %><br/>
```

Update `signature_count` to:

```ruby
def signature_count(key, count, options = {})
  options = options.merge({count: count, formatted_count: number_with_delimiter(count)})
  t(:"#{key}.html", scope: :"petitions.signature_counts", options)
end
```

h-lame: Might be easier for one of us to add this though as it probably impacts a test on `signature_count`.

PR #190, opened 2015-06-24
Style closed petition view

henryhadlow:
* Make buttons normal width on tablet/desktop
* Make a generic .secondary grey-text class
* Remove 'How petitions work' from petitions show
* Add new closed_petition_show view
* Move _petition_show to _open_petition_show and remove conditional logic
* Typography tweaks

### DIFF
--- a/app/views/petitions/_open_petition_show.html.erb
+++ b/app/views/petitions/_open_petition_show.html.erb
@@ -1,7 +1,3 @@
-<%= content_tag :p, :class => 'flash-notice' do -%>
-    This petition is now closed
-<% end if petition.closed? %>
-
 <h1>
   <span class="heading-secondary">Petition</span>
   <%= petition.action %>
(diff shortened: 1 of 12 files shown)


henryhadlow:
Hold it! Forgot to delete a redundant view

h-lame:
Assuming this goes green, :+1: from me.

app/views/petitions/_open_petition_show.html.erb
```diff
@@ -15,10 +11,22 @@
   </details>
 <% end %>
 
-<%# Signable #%>
-<% if petition.can_be_signed? -%>
```
alexander-bobin: Looks like `can_be_signed` is only being used in one other place now. It shouldn't be changed in this PR but we could probably drop it

alexander-bobin:
All looks OK to me accept I think the story needs some ACs in Pivotal so Alan can test it

PR #192, opened 2015-06-24
Increase local petitions to 50

yahoopete:


h-lame:
Note: reopened version of #189

PR #193, opened 2015-06-24
Add scoped search to listview pages

otlaitil:
For https://www.pivotaltracker.com/story/show/97497658

Removes search from front page and adds it to petition list views.


otlaitil:
This fixes also a template rendering issue discussed here: https://github.com/alphagov/e-petitions/commit/f823f4ef5e1b4c51aced1e43e397b0b8e5f6eb8d

features/step_definitions/moderation_steps.rb
```diff
@@ -49,7 +49,7 @@
 end
 
 Then /^the petition is not available for searching or viewing$/ do
-  step %{I search for "rejected" petitions with "#{@petition.action}"}
+  step %{I search for "Rejected" with "#{@petition.action}"}
   step %{I should not see the petition "#{@petition.action}"}
 end
```
h-lame: These two are inconsistent.  Should it be `I search for "Rejected" with "#{@petition.action}"` or `I search for "Rejected petitions" with "#{@petition.action}"}`, it's confusing if they're not the same.

features/suzie_searches_by_free_text.feature
```diff
@@ -16,11 +16,25 @@ Feature: Suzy Singer searches by free text
...
+
   Scenario: Search for open petitions
-    When I search for "Wombles"
-    Then I should be on the search results page
-    And I should see "Search results - Petitions" in the browser page title
-    And I should see /For "Wombles"/
+    When I go to the petitions page
+    And I follow "Open petitions"
+    And I fill in "Wombles" as my search term
+    And I press "Search"
```
h-lame: Nitpick: Could we use the `I search for "([^"]*)"? with "([^"]*)"` step to collapse these three steps?

app/helpers/search_helper.rb
```diff
@@ -9,4 +9,12 @@ def will_paginate_petitions(collection_or_options = nil, options = {})
        <span class='paginate paginate-next'>Next</span>"
     will_paginate *[collection_or_options, options].compact
   end
+
+  def filtered_petition_count(petitions)
+    if petitions.search?
+      pluralize(petitions.total_entries, 'result')
+    else
+      pluralize(petitions.total_entries, 'petition')
```
h-lame: Probably want a `number_with_delimiter` around the number (so we get `1,000,000` instead of `1000000`).

features/step_definitions/moderation_steps.rb
otlaitil: True. Will fix!

app/helpers/search_helper.rb
otlaitil: Yes - will fix!

otlaitil:
Thanks for comments @h-lame - I addressed them

h-lame:
Looks good. :+1:

davidbasalla:
:+1:

PR #195, opened 2015-06-24
Add animated progress bar/graph to petition show

henryhadlow:
Should this have tests?

### DIFF
--- a/app/assets/stylesheets/petitions/views/_petition-show.scss
+++ b/app/assets/stylesheets/petitions/views/_petition-show.scss
@@ -8,17 +8,48 @@
   }
 }
 
+// Signature count and signature graph
+
 .signature-count {
-  @include bold-48;
-  margin: em(10, 32) 0;
-  @include media(tablet) {
-    margin: em(15, 48) 0;
+  @extend %contain-floats;
+
+  .signature-count-number {
+    font-size: 36px;
+    line-height: 40/36;
+    font-weight: 700;
+    margin-top: em(15, 36);
+    margin-bottom: em(10, 36);
+
+    span {
+      font-size: 24px;
+    }
   }
 
-  span {
-    @include bold-36;
+  .signature-count-goal {
+    color: $secondary-text-colour;
+    text-align: right;
+    margin: 10px 0 0;
+  }
+
+  .signature-count-graph {
+    clear: both;
+    width: 100%;
+    height: 10px;
+    overflow: hidden;
+    background-color: $border-colour;
+    padding: 0;
+
+    .signature-count-current {
+      display: block;
+      height: 10px;
+      background-color: $petitions-brand;
+      @extend %animated;
+      @extend %fadeInLeft;
+    }
   }
 }
+
+// Petition meta
 
 .petition-meta {
   @extend %contain-floats;
--- a/app/views/petitions/_closed_petition_show.html.erb
+++ b/app/views/petitions/_closed_petition_show.html.erb
@@ -16,7 +16,27 @@
   <span>All petitions run for 6 months</span>
 </p>
 
-<p class="signature-count"><%= signature_count(:default, petition.signature_count) %></p>
+<div class="signature-count">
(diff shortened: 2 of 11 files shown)


h-lame:
Rebased to include commit for animated info block from #197 which was branched off this branch.

pixeltrix:
3200 lines of CSS for an animating bar? Seems a little excessive.

app/views/petitions/_closed_petition_show.html.erb
```diff
@@ -16,7 +16,27 @@
...
+      "></span>
+  </div>
+  <p class="signature-count-goal">
+    <% if petition.signature_count < Site.threshold_for_response %>
+      <%= number_with_delimiter(Site.threshold_for_response) %>
+    <% else %>
+      <%= number_with_delimiter(Site.threshold_for_debate) %>
+    <% end %>
+  </p>
+</div>
```
pixeltrix: This is too much logic to go in a template - it should be extracted to a helper method and tested

app/assets/stylesheets/petitions/views/_sign.scss
```diff
@@ -6,3 +6,7 @@
   padding-left: $gutter * 1.5;
   margin-bottom: $gutter * 1.3;
 }
+
+#signature_postcode, #petition_creator_signature_postcode {
+  text-transform: uppercase;
+}
```
pixeltrix: Are you having to do this because the data hasn't been sanitized properly? We should normalise the formatting of postcodes on the server.

alexander-bobin:
Agreed with @pixeltrix on '3200 lines of CSS for an animating bar? Seems a little excessive.'. Those `@keyframe` declarations will end up in the final css. They're not the same as @extend or @include where only what we use gets brought in.

app/assets/stylesheets/petitions/views/_petition-show.scss
```diff
@@ -8,14 +8,92 @@
...
+  }
+
+  100% {
+    opacity: 1;
+    -webkit-transform: none;
+    transform: none;
+  }
+}
+
+%animated {
```
alexander-bobin: How come we have %animated specified twice? Once in the source file and once here

henryhadlow: @pixeltrix @alexander-bobin – the animation only adds about 10 lines to the output css.
`%extend` syntax in scss isn't compiled unless called in a class. For example:
```.warning {
  @extend %animated;
  @extend %bounce;
}```
Thought this was a pragmatic balance but happy to just copy and paste the lines used into place.

henryhadlow: @alexander-bobin re %animated specified twice – yeah I should fix that

app/assets/stylesheets/petitions/views/_sign.scss
henryhadlow: We do have server-side validation. We the build team know this isn't case sensitive but plenty of users won't assume that. This is just a nicety for the user (I think it only affects how the `input` content is displayed).

app/views/petitions/_closed_petition_show.html.erb
henryhadlow: OK. I won't be able to do that myself. Can someone help?

app/assets/stylesheets/petitions/views/_petition-show.scss
alexander-bobin: Coolio. Better to copy and paste only what we need. If you look at the output you'll see lots of `@keyframe` declarations which do get included.

henryhadlow: OK, will take a look and rebase against master.

henryhadlow:
I'm still going to need help extracting the logic in `app/views/petitions/_closed_petition_show.html.erb` to a helper method

henryhadlow:
Holy moly it's green

alexander-bobin:
Merged manually

PR #196, opened 2015-06-24
Make controls on admin all petitions autosubmit filter form

h-lame:
For: https://www.pivotaltracker.com/story/show/97713774

When you choose a new state to filter the all petitions list by, or a new number of petitions to see per page, we want the user to not have to press go.  We do this with a simple jquery enhancement that attaches to any select element with a data-autosubmit attribute and adds a onchange handler to submit the form the select lives in.

We leave the 'Go' button just in case (but perhaps we should hide it?).

Notes of interest:

1. We test this only via cucumber but this means we had to change some of our inspection steps to use capybara matchers instead of rspec matchers.  Without using the capybara matcher we don't neccessarily wait for the page refresh so might assert on the previous page contents.

2. We have an old version of jquery so we need to do `$(this).closest('form').submit();` - a newer jquery would let us do `$(this).form().submit();`.

### DIFF
--- a/app/views/admin/petitions/index.html.erb
+++ b/app/views/admin/petitions/index.html.erb
@@ -1,8 +1,8 @@
 <%= form_tag admin_petitions_path, :method => :get, :class => "filter" do -%>
   <%= label_tag :state, "Filter by status:" %>
-  <%= select_tag :state, options_for_select(Petition::SELECTABLE_STATES, params[:state]), :include_blank => "-- all moderated petitions --" %>
+  <%= select_tag :state, options_for_select(Petition::SELECTABLE_STATES, params[:state]), :include_blank => "-- all moderated petitions --", data: {autosubmit: true} %>
   <%= label_tag :per_page, "Results per page:" %>
-  <%= select_tag :per_page, options_for_select(%w(20 50 100), params[:per_page]) %>
+  <%= select_tag :per_page, options_for_select(%w(20 50 100), params[:per_page]), data: {autosubmit: true} %>
   <%= submit_tag "Go" %>
 <% end -%>
 <table class="admin_index">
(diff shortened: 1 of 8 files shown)


yahoopete:
I don't think we will need the custom pagination option any longer if that
makes any difference.

h-lame:
@yahoopete it doesn't make much difference, but if we don't want it I can delete it and simplify some specs / backend code by about 1%.  Shall I?

yahoopete:
The story to strip the pagination options is coming this week, maybe we
could play it earlier if easier?

https://www.pivotaltracker.com/story/show/97713802

Not bothered either way.

h-lame:
Let's do it in that story then.

app/views/admin/petitions/index.html.erb
```diff
@@ -1,8 +1,8 @@
 <%= form_tag admin_petitions_path, :method => :get, :class => "filter" do -%>
   <%= label_tag :state, "Filter by status:" %>
-  <%= select_tag :state, options_for_select(Petition::SELECTABLE_STATES, params[:state]), :include_blank => "-- all moderated petitions --" %>
+  <%= select_tag :state, options_for_select(Petition::SELECTABLE_STATES, params[:state]), :include_blank => "-- all moderated petitions --", data: {autosubmit: true} %>
   <%= label_tag :per_page, "Results per page:" %>
-  <%= select_tag :per_page, options_for_select(%w(20 50 100), params[:per_page]) %>
+  <%= select_tag :per_page, options_for_select(%w(20 50 100), params[:per_page]), data: {autosubmit: true} %>
```
otlaitil: Nice approach!

otlaitil:
LGTM :+1:

davidbasalla:
Cool!

PR #198, opened 2015-06-25
Drop email signees

h-lame:
For: https://www.pivotaltracker.com/story/show/97008326

Now that edit_response only does one thing, we don't need to have the email_signees checkbox because the intention is that you will always email the signatories when you change the response.  We also take the opportunity to tidy up some copy and standardize on the email button text etc...


davidbasalla:
looks good

alexander-bobin:
LGTM :shipit: 

@h-lame, needs a rebase. Lemme know when done and I'll test locally

app/views/admin/debate_outcomes/show.html.erb
```diff
@@ -29,6 +29,6 @@
   <% end %>
 
   <%= form_row do %>
-    <%= f.submit 'Publish & Send Email', data: { confirm: "Are you sure you want to publish these debate outcome details and email all of the supporters?" } %>
+    <%= f.submit "Email #{number_with_delimiter(@petition.signature_count)} #{'signature'.pluralize}" %>
```
otlaitil: It might be good to extract this pluralizing to a helper, the pattern is used in at least 4 places now.

otlaitil:
LGTM :+1:

app/views/admin/debate_outcomes/show.html.erb
h-lame: Yeah, that's a good point.  Will do!

h-lame:
ok, rebased - if we're green merge it

alexander-bobin:
Tests pass locally. Merges fine locally. Will merge now.

PR #200, opened 2015-06-25
Whitelist facets exposed on search pages

h-lame:
For: https://www.pivotaltracker.com/story/show/97724776

To give us control over what search facets are accessible on the frontend vs backend we define a list of facets in the locale file under `petitions.facets.public` and `petitions.facets.admin`.  The public frontend only shows these facets in the filter list so a user shouldn't be invited to click on a facet that isn't public.  We also make the public backend reject any facet that is not in the public list so they can't just hack the url to see, for example, a list of hidden petitions.

The admin isn't so constrained as it's not a problem if an admin can see a petition via a non-public scope.  However, so we don't baffle them with too many options we use the locale list to populate the filter dropdown on the all petitions page.

### DIFF
--- a/app/helpers/petition_helper.rb
+++ b/app/helpers/petition_helper.rb
@@ -21,15 +21,7 @@ module PetitionHelper
   end
 
   def public_petition_facets_with_counts(petition_search)
-    Hash[
-      public_petition_facets.map do |key|
-        if petition_search.facets.key? key.to_sym
-          [key, petition_search.facets[key.to_sym]]
-        else
-          nil
-        end
-      end.compact
-    ]
+    petition_search.facets.slice(*public_petition_facets).stringify_keys
   end
 
   private
--- a/app/helpers/petition_helper.rb
+++ b/app/helpers/petition_helper.rb
@@ -22,7 +22,7 @@ module PetitionHelper
 
   def public_petition_facets_with_counts(petition_search)
     Hash[
-      I18n.t('public', scope: :"petitions.facets").map do |key|
+      public_petition_facets.map do |key|
         if petition_search.facets.key? key.to_sym
           [key, petition_search.facets[key.to_sym]]
         else
--- a/app/helpers/petition_helper.rb
+++ b/app/helpers/petition_helper.rb
@@ -20,6 +20,18 @@ module PetitionHelper
     t(:"petitions.rejection_reasons.descriptions")
   end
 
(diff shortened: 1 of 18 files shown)


h-lame:
Obvs, props to @pixeltrix as this was his idea, I'm just the crude workman who implemented it.

otlaitil:
Feels a bit strange to use a localisation file to configure the application... But OK :) Implementation looks good. :+1:

features/admin/all_petitions.feature
```diff
@@ -37,19 +37,19 @@ Feature: A moderator user views all petitions
...
 
-    And I filter the list to show "rejected" petitions
+    And I filter the list to show "Rejected" petitions
     Then I should see the following list of petitions:
      | My rejected petition |
 
-    And I filter the list to show "hidden" petitions
+    And I filter the list to show "Hidden" petitions
     Then I should see the following list of petitions:
      | My hidden petition   |
```
davidbasalla: I have added a bunch of tests for the newly available lists (awaiting debate, etc) in my branch, will add them once this gets merged!

davidbasalla:
Looks good :shipit: once :green_apple: (so I can merge it into my branch :) )

davidbasalla:
Needs a rebase :/

app/helpers/petition_helper.rb
```diff
@@ -32,6 +32,18 @@ def waiting_for_debate_date_in_words(petition)
...
+    Hash[
+      public_petition_facets.map do |key|
+        if petition_search.facets.key? key.to_sym
+          [key, petition_search.facets[key.to_sym]]
+        else
+          nil
+        end
+      end.compact
+    ]
+  end
```
pixeltrix: If you implement `slice` on `Browseable::Facets` you can simplify this down, e.g:

``` ruby
module Browseable
  class Facets
    def slice(*keys)
      keys = keys.map(&:to_sym)

      keys.each_with_object({}) do |key, hash|
        hash[keys] = self[key] if has_key?(key)
      end
    end
  end
end

@petitions.facets.slice(public_facets)
```

h-lame:
Rebased and addressed @pixeltrix's comments.  Merge plz.

davidbasalla:
Wasn't fast enough with the merge, rebase needed

h-lame:
2nd time (today)'s the charm?

PR #204, opened 2015-06-25
Add background job to verify petition signature counts

pixeltrix:
Since we are now calculating the signature counts relatively rather than using a cron job we should be prudent and add a background job that regularly checks for invalid signature counts and then if it finds any to notify NewRelic of the ones that have occurred.

https://www.pivotaltracker.com/story/show/97710094


otlaitil:
:+1:

h-lame:
:+1: (obvs as I've already merged it)

PR #205, opened 2015-06-25
Remove Petition.states_by_count

pixeltrix:
The method Petition.states_by_count was used for the counts in tabs and is no longer used anywhere so remove it.


davidbasalla:
cool

h-lame:
Yay, deletes!

PR #210, opened 2015-06-26
Style admin hub

alexander-bobin:
https://www.pivotaltracker.com/story/show/97713360

- Create hub page and integration tests
- Add generic styling

Not done:
- link to debate queue because we don't have that page yet
- icons in the boxes because we don't have them yet
- extensive feature tests for the links to listing pages because the pages don't exist yet


app/assets/stylesheets/petitions/admin/_temporary_tab_nav.scss
```diff
@@ -0,0 +1,16 @@
+.admin {
```
otlaitil: Why all the rules here wrapped in `.admin`, the child element seems to be referred with id and the `.admin` itself does not have any styles?

alexander-bobin: It's a small namespace for any instance where we are overwriting public styles. This may change but it helps reduce the likelihood someone will make a change that will effect the public site.

otlaitil: Ok! Good to know.

app/views/admin/admin/index.html.erb
```diff
@@ -0,0 +1,16 @@
+<h1>Petition admin</h1>
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <% action_counts do |counts| %>
+      <ul class="hub-actions grid-row">
+        <% counts.each do |state, count| %>
+          <li class="column-half action-<%= state.to_s.gsub('_', '-') %>">
```
otlaitil: `<%= state.to_s.gsub('_', '-') %>` I wonder if we could refactor this out of the view. Maybe use the locale file to pull out the string with the state as a key, or add another variable 'css_class' to `generate_counts` and use it like

`counts.each do |state, css_class, count|`

otlaitil:
Just one minor thing (see the comment). LGTM :+1:

app/views/admin/admin/index.html.erb
h-lame: activesupport gives us `state.dasherize`

otlaitil: Cool, that's much better.

PR #211, opened 2015-06-26
Remove per-page UI and use 50 in all paginated lists 50 in admin

h-lame:
For: https://www.pivotaltracker.com/story/show/97713802

### DIFF
--- a/app/controllers/admin/petitions_controller.rb
+++ b/app/controllers/admin/petitions_controller.rb
@@ -2,7 +2,7 @@ class Admin::PetitionsController < Admin::AdminController
   respond_to :html
 
   def index
-    @petitions = Petition.selectable.search(params.merge(count: params[:per_page] || 20))
+    @petitions = Petition.selectable.search(params.merge(count: 50))
   end
 
   def show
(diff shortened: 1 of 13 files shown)


app/controllers/admin/petitions_controller.rb
```diff
@@ -10,7 +10,7 @@ def show
   end
 
   def threshold
-    @petitions = Petition.threshold.order(:signature_count).paginate(:page => params[:page], :per_page => params[:per_page] || 20)
+    @petitions = Petition.threshold.order(:signature_count).paginate(:page => params[:page], :per_page => 50)
```
otlaitil: Would we gain anything by always using our custom search? It includes the pagination and the implementation would be uniform (`@petitions` would always be an instance of `Browseable:Search` everywhere - now it's mixed with `ActiveRecord:Relation`).

otlaitil:
LGTM :+1:

app/controllers/admin/petitions_controller.rb
davidbasalla: What Otto said, although this method will disappear once my branch https://github.com/alphagov/e-petitions/pull/212 gets merged in (removing threshold and todolist).

mybits:
LGTM :+1:

davidbasalla:
looks good

PR #213, opened 2015-06-26
Change closed to a real state

pixeltrix:
As part of a series of steps to simplify our list queries make the closed state a real state and not based on it being open and the closed_at time being in the past.

### DIFF
--- a/app/models/petition.rb
+++ b/app/models/petition.rb
@@ -57,7 +57,7 @@ class Petition < ActiveRecord::Base
   # = Validations =
   include Staged::Validations::PetitionDetails
   validates :response_summary, length: { maximum: 500, message: 'Response summary is too long.' }
-  validates_presence_of :open_at, :closed_at, :if => :open?
+  validates_presence_of :open_at, :if => :open?
   validates_presence_of :rejection_code, :if => :rejected?
   validates_inclusion_of :rejection_code, :in => REJECTION_CODES, :if => :rejected?
   # Note: we only validate creator_signature on create since if we always load creator_signature on validation then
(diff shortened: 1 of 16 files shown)


mybits:
:+1: for simplifying things

h-lame:
``Petition#open?`` is no longer symmetrical with ``Petition#closed?`` depending on the time you run them.  Something can be ``open? #=> true`` and ``closed? #=> true`` if the in-memory instance crossed the ``closed_at`` boundary.  It's minor, sure, but it does mean we've got some danger in how we write our views and controller actions if we want to guard against folk loading a page just before midnight and submitting a signature just after.

There is ``Petition#can_be_signed?`` which has the logic we want, maybe we should merge that with ``#open?``.  Or get rid of it.

Separately, but related, our scopes for fetching open and closed petitions don't care about ``closed_at`` but our instance methods do.  It's fine (we've done it for a performance reasons) but we maybe want to leave a note somewhere in the code.

app/models/petition.rb
```diff
@@ -229,8 +231,8 @@ def hidden?
     state == HIDDEN_STATE
   end
 
-  def closed?
-    state == OPEN_STATE && closed_at <= Time.current
+  def closed?(time = Time.current)
```
h-lame: Total nitpick: but we don't need this arg anymore.

lib/tasks/import.rake
```diff
@@ -35,7 +35,7 @@ namespace :import do
         petition.title              = row['title']
         petition.description        = row['description']
         petition.response           = row['response']
-        petition.state              = row['state']
+        petition.state              = row['state'] == 'open' ? 'closed' : row['state']
```
h-lame: This is just brilliant, not confusing at all ;)

spec/models/petition_spec.rb
```diff
@@ -474,117 +468,173 @@
...
+        expect(petition.closed?).to be_truthy
+      end
+    end
+
+    context "when the closed_at has not passed but the state is closed" do
+      let(:petition) { FactoryGirl.build(:petition, state: Petition::CLOSED_STATE, closed_at: 3.days.from_now) }
+
+      it "return true" do
+        expect(petition.closed?).to be_truthy
+      end
```
h-lame: Can we just collapse these down to "is closed when state is closed"? like the others?  I admit that I find the title of `"when the closed_at has passed but the state is still open"` a bit confusing (compared to the implementation which uses `CLOSED_STATE`, not `OPEN_STATE`.

PR #216, opened 2015-06-28
Refactor page titles and content for

pixeltrix:
Using `content_for` inside nested templates makes caching tricky because those block will not get called if those templates are cached. This PR eliminates that usage and makes some changes to trending petitions as well to so that it can be cached in the view.


h-lame:
Looks good, needs a rebase tho'

pixeltrix:
Rebased, merged when :green_heart:

PR #220, opened 2015-06-28
Many Admin improvements

mybits:
PT: https://www.pivotaltracker.com/n/projects/307301/stories/97713112

we want to get rid of `search_controller` from admin and move the logic to the index action in `petitions_controller`.


h-lame:
Also included:

PT: https://www.pivotaltracker.com/story/show/97990386

remove redundant single link tabnav (we remove the gem entirely as the only other use was in the public site and didn't need to be dynamic, so writing out the HTML was easier).

PT: https://www.pivotaltracker.com/story/show/97990370

Add search to the main admin hub too

PT: https://www.pivotaltracker.com/story/show/97827046

Add a "in debate queue" admin hub blob + filter for all petitions.  Note that this uses the timestamp rather than the petition count - so it's that a petition reached the threshold or it has been scheduled, not the number of signatures (although the number if signatures is what triggers the timestamp so...)

In passing we also tidy up a bunch of things in the admin to remove useless code.

pixeltrix:
Rebased and merged in 676daf7f2a77d5752befd888b64475c2ee7b64de

PR #224, opened 2015-06-29
Remove character count when checking for existing petitions

alexander-bobin:
Drop the character counter (as it is a guide only) and truncate action
at 255 instead of 80


alanth:
Nice and simple, putting this through

PR #229, opened 2015-06-30
Bugfix to mail threshold, not signature count

yahoopete:
PT: https://www.pivotaltracker.com/story/show/98080170


yahoopete:
celebrate

PR #230, opened 2015-06-30
Extract formatted threshold numbers to the site model

pixeltrix:
We are calling these formatted numbers in a number of places so add some convenience methods to make generating these easier.


h-lame:
Oh, yes, good call.

PR #231, opened 2015-06-30
Use id lookup in SignaturesController#signed

h-lame:
This page, the final step in signing a petition, had a url ``/petitions/<id>/signatures/<token>/signed`` which meant to find the signature we did a full table scan lookup on ``Signature#perishable_token``.  This is not particularly performant, particularly as ``perishable_token`` is not indexed, as such it was a bottleneck in the signing flow: these slow full table scans cause a lock on any writes to the table (e.g. for other signers at the initial signing or verification point in the flow).

The fix is to rework this action to be like ``SignaturesController#verify`` and ``SignaturesController#unsubscribe`` and lookup by id, but reject the request if the token does not match.  This means moving the url from ``/petitions/<id>/signatures/<token>/signed`` to ``/signatures/<id>/signed/<token>``.

(Discovered as part of: https://www.pivotaltracker.com/story/show/95979302)

### DIFF
--- a/app/controllers/signatures_controller.rb
+++ b/app/controllers/signatures_controller.rb
@@ -2,7 +2,7 @@ class SignaturesController < ApplicationController
   include ManagingMoveParameter
 
   before_filter :retrieve_petition, :only => [:new, :create, :thank_you]
-  before_filter :retrieve_signature, :only => [:verify, :unsubscribe]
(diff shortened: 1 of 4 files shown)


oskarpearson:
Seems like a better idea to me - thanks!

PR #232, opened 2015-06-30
Style error pages

henryhadlow:
Move styles inline on error pages
Move images to public (outside asset pipeline)
Remove maintenance.html and overloaded.html


pixeltrix:
I think you can split out the CSS to a static file in `public`

alexander-bobin:
Shame we had to include the images. BUT, LGTM

PR #233, opened 2015-06-30
Style admin petition show and actions

henryhadlow:
This is as far as I've got with styling. @h-lame or @pixeltrix to take a look at the best way to get the mechanics working.

Prototype https://marvelapp.com/e08dab#6713178 shows how I want it to work (**but doesn't mean the form fields have to be inline**)

You might want to use the first commit only – see what you think.


app/assets/stylesheets/petitions/_layout.scss
```diff
@@ -74,3 +74,7 @@
 .column-two-thirds {
   @include grid-column(2/3);
 }
+
```
alexander-bobin: 2 things:

- If this is only used in admin, let's put it into an admin specific stylesheet like admin/_layout.scss
- The class name describes what it does rather than what it is. Is there another name we can use like `.admin-main` or something?

app/assets/stylesheets/petitions/admin/_actions.scss
```diff
@@ -0,0 +1,17 @@
+nav.actions {
```
alexander-bobin: Nitpick: Don't need the `nav` node (or `li` below) unless we are specifically upping specificty

app/assets/stylesheets/petitions/admin/_meta.scss
```diff
@@ -0,0 +1,7 @@
+.admin-petition-meta {
```
alexander-bobin: FYI: In other admin scss files we are using namespacing to make it doubly specific. This is OK though cause it is clearly an admin class in an admin folder

app/views/admin/petitions/_petition_details.html.erb
```diff
@@ -1,33 +1,33 @@
-<div class="row">
-  <label>Petition Id:</label>
-  <p class="text_block"><%=  @petition.id %></p>
-</div>
+<h2>Status</h2>
```
alexander-bobin: Nitpick: Could have used a `<dl>` for this.

alexander-bobin:
Couple of small points. I think the only thing that should be done is moving `.extra-gutter` to an admin stylesheet.

h-lame:
Did some stuff, hopefully addressed @alexander-bobin's comments too.  @henryhadlow not sure if there's more styling you wanted to add?

henryhadlow:
@h-lame nice one. Yup, I'll jump in there in a mo / after lunch.

alexander-bobin:
LGTM. Will wait for @henryhadlow, to look over first

PR #234, opened 2015-06-30
Restructure indexes

oskarpearson:
Ideally we'd do performance tests of each change here. Hopefully this is a net win.

See individual commits for more detailed comments.
- Changed the order of columns in some of the indexes
- Removed some unnecessary indexes for performance


h-lame:
No reason not to merge this.

PR #235, opened 2015-06-30
Slimming transactional queries for validate

h-lame:
@oskarpearson had a look at our query log during performance testing and noticed that doing signature validation had a bunch of queries in a transaction.  This PR is about trying to reduce that number:

1. move creator signature validation out of the transaction - it's a bit belt-and-braces that we even have it in here still, as the creator signature should be validated earlier in the process.  It doesn't hurt to try (as it's almost a no-op if it's already validated), but it doesn't need to happen in the transaction for validating the current signature
2. reduce the queries used by stamping the constituency petition journal.
  1. when the record already exists don't do a reload to refresh the in-memoery object after we issue the atomic sql `UPDATE` statement
  2. when the record doesn't exist, don't save it and then update it (2 queries), just insert with the correct signature count (1 query)

### DIFF
--- a/app/models/constituency_petition_journal.rb
+++ b/app/models/constituency_petition_journal.rb
@@ -15,7 +15,7 @@ class ConstituencyPetitionJournal < ActiveRecord::Base
   }
 
   def self.for(petition, constituency_id)
-    find_or_create_by(petition: petition, constituency_id: constituency_id)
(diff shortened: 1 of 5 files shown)


h-lame:
Merging after verbal "ok" from @pixeltrix.

PR #237, opened 2015-06-30
Synthesize closing at information for petitions

h-lame:
Since merging in #213 we no longer have a `closed_at` timestamp filled in on open petitions so we can't show a deadline.  We implement `Petition#deadline` to work around this by advancing ahead from `open_at` by the appropriate number of months and using this new method wherever we were using `.closed_at`.

### DIFF
--- a/app/models/petition.rb
+++ b/app/models/petition.rb
@@ -243,6 +243,14 @@ class Petition < ActiveRecord::Base
     TODO_LIST_STATES.include? state
   end
 
+  def deadline
+    if closed?
+      closed_at
+    else
+      Site.closed_at_for_opening(open_at)
+    end
+  end
+
   def rejection_reason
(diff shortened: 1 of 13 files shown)


app/models/petition.rb
```diff
@@ -247,6 +247,16 @@ def state_label
...
+  def deadline
+    if closed?
+      closed_at
+    elsif open_at?
+      Site.closed_at_for_opening(open_at)
+    else
+      nil
+    end
+  end
+
```
pixeltrix: Can be implemented this way:

```
def deadline
  open_at && (closed_at || Site.closed_at_for_opening(open_at)
end
```

pixeltrix:
Other than the one comment :shipit:

h-lame:
Rebased and addressed @pixeltrix's comment.  Merge when green please.

PR #238, opened 2015-07-01
Redirect admin users to the hub when they sign in

yahoopete:
https://www.pivotaltracker.com/story/show/98122274

This is where their in-tray lives.  Let's do that.


alexander-bobin:
LGTM

PR #239, opened 2015-07-01
Redirect back to the petition show page in admin after adding a response.

yahoopete:
PT: https://www.pivotaltracker.com/story/show/98117102


alexander-bobin:
LGTM

PR #242, opened 2015-07-01
Drop the strapline like it's hot

yahoopete:
https://www.pivotaltracker.com/story/show/98178604


alanth:
Never like that strapline anyway. Merging.

PR #243, opened 2015-07-01
Optimise signature number

oskarpearson:
The 'signature number count' would previously do a very long running select
(1 second runtime per page display on a big db).

We now store the signature sequence in each recorded signature, so that it can be
displayed to the user in a consistent way (even if the page is reloaded).

The signature number is populated based on the current Petition's signature
count at the time of signing.

Note that we've taken the decision that if two people sign at exactly the same
time, they could both be signature number 3 (it's a "dead heat" situation).
This way we don't need to grab a lock on Petitions table before doing the
increment. However, the next person signing would jump to signature number 5,
since the increment of the Petitions table _is_ done in a transaction.

(Discussed initially with Andrew White)

### DIFF
--- a/app/models/signature.rb
+++ b/app/models/signature.rb
@@ -64,18 +64,8 @@ class Signature < ActiveRecord::Base
   # = Methods =
   attr_accessor :uk_citizenship
 
-  def self.signature_number(petition_id, validated_at)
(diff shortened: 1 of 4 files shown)


h-lame:
I'm into it.

PR #246, opened 2015-07-01
Add some JS to show / hide rejection interface when moderating

h-lame:
There's no point seeing the rejection interface if you choose approve, and there's no point seeing it before you decide what to do.  We do always show it if there's an error on the object though.

The impl is just some inline JS because we already use inline JS for the rejection description stuff in the rejection interface.  Even if this isn't great, at least it's consistent.


h-lame:
Did try to do this with CSS only as we thought adjancy and attribute seletors might work, but the elements don't lend themselves to that sort of selector.

PR #247, opened 2015-07-01
Change 'Lists of petitions' to use a details elements

henryhadlow:


henryhadlow:
Not sure how to fix these tests or why they are failing :unamused:

yahoopete:
The tests are looking finding 2 instances of things like Rejected petition so don't know ow which one to select. 

It's messy.  X path constraints....  I had to hack a test script helper to tell it where to look for a similar problem. 

When I say "I",  I mean @h-lame

h-lame:
Taking a look.

h-lame:
Did a thing, should be green and mergable now.

alexander-bobin:
LGTM

PR #248, opened 2015-07-01
Show correct error message when no constituency petitions found

h-lame:
Because we were using ``@petitions.present?`` as our guard and both ``[].present?`` and ``.present?`` on an AR relation with no results return ``false`` we ended up telling the user that we knew their MP but didn't know their constituency.  We switch to using ``@constituency.present?`` as the guard and this restores the correct behaviour.


pixeltrix:
LGTM

PR #249, opened 2015-07-01
Refactor models

pixeltrix:
Break out rejection, response and notes to separate models and refactor scopes and facets to be simple queries. Needs a follow up to optimise the indexes for the queries.


pixeltrix:
After these changes the facet queries are as follows:

```
SELECT COUNT(*) FROM "petitions" WHERE "petitions"."state" IN ('open', 'closed', 'rejected')
SELECT COUNT(*) FROM "petitions" WHERE "petitions"."state" IN ('open', 'closed', 'rejected') AND "petitions"."state" = $1  [["state", "open"]]
SELECT COUNT(*) FROM "petitions" WHERE "petitions"."state" IN ('open', 'closed', 'rejected') AND "petitions"."state" = $1  [["state", "closed"]]
SELECT COUNT(*) FROM "petitions" WHERE "petitions"."state" IN ('open', 'closed', 'rejected') AND "petitions"."state" = $1  [["state", "rejected"]]
SELECT COUNT(*) FROM "petitions" WHERE "petitions"."state" IN ('open', 'closed', 'rejected') AND ("petitions"."response_threshold_reached_at" IS NOT NULL) AND "petitions"."government_response_at" IS NULL
SELECT COUNT(*) FROM "petitions" WHERE "petitions"."state" IN ('open', 'closed', 'rejected') AND ("petitions"."government_response_at" IS NOT NULL)
SELECT COUNT(*) FROM "petitions" WHERE "petitions"."state" IN ('open', 'closed', 'rejected') AND ("petitions"."debate_threshold_reached_at" IS NOT NULL) AND "petitions"."scheduled_debate_date" IS NULL
SELECT COUNT(*) FROM "petitions" WHERE "petitions"."state" IN ('open', 'closed', 'rejected') AND ("petitions"."debate_outcome_at" IS NOT NULL)
```

The lolsql bit where we have two conditions on `state` is because we're scoping it to `Petition.visible` and then doing the query to ensure that only published petitions are shown. I'm pretty sure the DB will deal with that in the query planner - I remember we took some code out of Rails to try and detect theses because it wasn't worth the effort.

alexander-bobin:
HTML LGTM. Really like the reorganisation of the facets and scopes. The rest I can't comment on

app/views/admin/petitions/_petition.html.erb
```diff
@@ -7,5 +7,5 @@
   </td>
   <td class="count"><%= number_with_delimiter(petition.signature_count) %></td>
   <td><%= date_format_admin(petition.deadline) %></td>
-  <td class="center"><%= petition.response.present? ? image_tag('admin/tick.gif') : '' %></td>
+  <td class="center"><%= petition.government_responded? ? image_tag('admin/tick.gif') : '' %></td>
```
h-lame: We just removed this method, but the tests are passing so I guess this partial maybe isn't being used anymore?

h-lame:
99% - I think we just need to remove a partial and we're good to merge.

pixeltrix:
Partial removed - good to go when it's green

PR #250, opened 2015-07-01
Add HttpOnly to the seen_cookie_message cookie

pixeltrix:
Although there was no way of exploiting this for evil it's best practise not to allow JavaScript access to cookies unless necessary for the operation of the application.

Fixes: GDNT-025-3-3


h-lame:
Hard to argue with.

h-lame:
However, didn't we take this code from somewhere else in the GDS stack?  Should we PR back upstream so everyone avoids this security hole in the future?

PR #252, opened 2015-07-02
Setup memcache

pixeltrix:
Just installing the gem and a couple of simple things first to check that it's working on the staging and preview environments.


oskarpearson:
Looks like a good change to me. And once it's in we can start caching other things

h-lame:
Looks good.

PR #265, opened 2015-07-03
Tagged searches via notes in admin

h-lame:
For: https://www.pivotaltracker.com/story/show/98300716

Doing a search for `[<something>]` will look only in admin notes fields for `[<something>]`.  This allows admins to "tag" petitions and find them later.  It's an exact match using a SQL `LIKE` statement, and we sanitize it to assume that `<something>` is a single "tag". (e.g. spaces are allowed in tag names and we strip extra `[]` from the search so that `[foo] [bar]` will end up as a search for `[foo bar]`.

### DIFF
--- a/app/controllers/admin/petitions_controller.rb
+++ b/app/controllers/admin/petitions_controller.rb
@@ -2,8 +2,11 @@ class Admin::PetitionsController < Admin::AdminController
   respond_to :html
 
   def index
-    @petitions = Petition.search(params.merge(count: 50))
-    @query = params[:q]
+    if filter_by_tag?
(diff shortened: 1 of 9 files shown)


h-lame:
If we had more time a full tagging system using something like `acts_as_taggable_on` or whatever might be a better, more manageable, solution.

yahoopete:
Love the minimal viability!

:+1:

pixeltrix:
Other that minor comments :+1:

pixeltrix:
Merged in b670dae9fb4e7ae69d6a251a85fee34cb7659516

PR #267, opened 2015-07-03
Styling the admin search, adding placeholder.

yahoopete:


app/views/admin/searches/_form.html.erb
```diff
@@ -1,5 +1,5 @@
-<%= form_tag [:admin, :search], method: :get do %>
-  <%= label_tag :q, "Search for id, title or an email address:" %>
-  <%= text_field_tag :q, query %>
-  <%= submit_tag 'Search', name: nil %>
+<%= form_tag [:admin, :search], class: 'search-inline', method: 'get' do %>
+  <%= label_tag :q, "Search all petitions" %>
+  <%= search_field_tag 'q', @petitions.query, class: 'form-control', id: 'search', placeholder:"Keywords, email address, petition ID or [tag]" %>
```
h-lame: This should just be `query`.  We pass in the `query` param to this partial, and not every page has an ``@petitions`` instance variable, or an ``@petitions`` object that has a `query`

PR #271, opened 2015-07-03
Fix discrepancy between model and db for Petition#background

h-lame:
The model validates at 300 chars, but the db was limited to 200.  Set both to 300.  We also add ``shoulda-matcher`` specs for the db fields of the model which raises some other discrepancies: we validate the presence of ``background`` and ``additional_details``, but the db columns are nullable.  These are less likely to break though as it's the model that is more restrictive than the db.

For: https://www.pivotaltracker.com/story/show/98300682


alanth:
:+1:

PR #273, opened 2015-07-03
Add routing constraints

pixeltrix:
Prevents public urls being accessed on the admin domain and vice-versa - see commits for gory details.


h-lame:
such yak.  many shave.  wow.

This all seems good to me (if annoying that we have to jump through these hoops).  All I think we need is to add something to the readme or comment header on relevant files (the proxy for example) because it is a bit weird and a short pointer to why wouldn't go amiss.

h-lame:
Happy to have this merged when rebased.

pixeltrix:
Please merge when green

PR #275, opened 2015-07-05
Replace NewRelic with AppSignal

pixeltrix:
Due to data protection concerns we can't use NewRelic for production so replace it for application monitoring with AppSignal which is based in the Netherlands and so is acceptable because it's in the EU. AppSignal only does errors and application monitoring but we can use the AWS CloudWatch service for the server monitoring side of things.

Needs to be synchronised with changes to the AWS stack before merging.


h-lame:
Assuming the deploy stuff is setup to push the keys and config out, I see no reason not to merge this when rebased.

pixeltrix:

PR #276, opened 2015-07-05
Compile assets in place on the server

pixeltrix:
To cover the period between when a new version of the app is deployed and the cached version expires then we need to keep the old version around. By packaging those up on deployment they get deleted by the CodeDeploy agent so create them in place on the server into a shared assets folder that's symlinked into the current release's public folder.

Needs to be synchronised with the ansible changes to the stack before merging


oskarpearson:
Looks good to me.

PR #277, opened 2015-07-05
Add subdomains to the HSTS headers

pixeltrix:
Rails doesn't include subdomains by default when sending the HSTS header as it doesn't know whether this is true but we do so force it to be true.


oskarpearson:
LGTM

pixeltrix:

PR #281, opened 2015-07-06
Homepage improvements

henryhadlow:
* Make threshold blocks clickable
* Add hover state for inline search submit
* Apply pull-quote style to all 3 government responses

### DIFF
--- a/app/assets/stylesheets/petitions/_search-inline.scss
+++ b/app/assets/stylesheets/petitions/_search-inline.scss
@@ -30,5 +30,8 @@
     background-repeat: no-repeat;
     background-position: 45% 50%;
     @include background-image("icons/icon_search-white", 18px, 18px);
+    &:hover, &:focus {
+      background-color: darken($petitions-panel-colour, 5%);
+    }
   }
 }
--- a/app/views/pages/index.html.erb
+++ b/app/views/pages/index.html.erb
@@ -29,20 +29,34 @@
 
 <% explanation_petitions do |actioned| %>
   <section aria-labelledby="response-threshold-heading" id="petitions-with-response">
-    <div class="threshold-panel response-threshold">
-      <h2 id="response-threshold-heading"><%= Site.formatted_threshold_for_response %></h2>
-      <p>If a petition gets <%= Site.formatted_threshold_for_response %> signatures, the government will&nbsp;respond</p>
-    </div>
+    <% if actioned[:with_response][:count].zero? %>
+      <div class="threshold-panel response-threshold">
+        <h2 id="response-threshold-heading"><%= Site.formatted_threshold_for_response %></h2>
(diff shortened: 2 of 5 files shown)


app/views/pages/index.html.erb
```diff
@@ -29,20 +29,24 @@
 
 <% explanation_petitions do |actioned| %>
   <section aria-labelledby="response-threshold-heading" id="petitions-with-response">
-    <div class="threshold-panel response-threshold">
-      <h2 id="response-threshold-heading"><%= Site.formatted_threshold_for_response %></h2>
-      <p>If a petition gets <%= Site.formatted_threshold_for_response %> signatures, the government will&nbsp;respond</p>
-    </div>
+    <%= link_to( petitions_path(state: :with_response) ) do %>
```
alexander-bobin: This needs some more attention. If we are wrapping the threshold panels with a link, then the link should become the threshold panel. This will mean that we apply the hover and focus bg colour changes to the link, not the div

app/assets/stylesheets/petitions/_search-inline.scss
```diff
@@ -30,5 +30,8 @@
     background-repeat: no-repeat;
     background-position: 45% 50%;
     @include background-image("icons/icon_search-white", 18px, 18px);
+    &:hover, &:focus {
```
alexander-bobin: Would prefer not to use the & scss pattern. It visually implies a nested element.

Better to just use a flat `.thing:hover`

henryhadlow:
Nice one Alex. I'll take a look and fix the tests ;)

henryhadlow:
Hold it! Gonna make some additional tweaks here

app/views/pages/index.html.erb
```diff
@@ -29,20 +29,34 @@
 
 <% explanation_petitions do |actioned| %>
   <section aria-labelledby="response-threshold-heading" id="petitions-with-response">
-    <div class="threshold-panel response-threshold">
-      <h2 id="response-threshold-heading"><%= Site.formatted_threshold_for_response %></h2>
-      <p>If a petition gets <%= Site.formatted_threshold_for_response %> signatures, the government will&nbsp;respond</p>
-    </div>
+    <% if actioned[:with_response][:count].zero? %>
+      <div class="threshold-panel response-threshold">
+        <h2 id="response-threshold-heading"><%= Site.formatted_threshold_for_response %></h2>
```
alexander-bobin: Nit pick: shame about the repetition of the h2 and p.

alexander-bobin:
LGTM

app/views/pages/index.html.erb
henryhadlow: I know :(
I wanted to use `link_to_unless` but it doesn't accept the url and then the `a` tag contents in a `do end` block

PR #284, opened 2015-07-06
Fix cacheable SSL responses

pixeltrix:
Although most of the website is acceptable for caching like browsing the home page, petition list pages and the petition show page, there are some sensistive pages like petition creation, sponsoring and sigining petitions. Also the admin should not be cached either.

Fixes GDNT-025-3-4.


oskarpearson:
Looks good to me. :+1:

pixeltrix:

PR #291, opened 2015-07-07
Fix bug where `action` was output in support flow `<title>`

henryhadlow:
Fix bug where the petition `action` was being displayed in the Support petition flow `<title>` tag


alanth:
:+1;

PR #292, opened 2015-07-08
Improve homepage when there are no petitions

henryhadlow:
Couldn't sleep :/

### DIFF
--- a/app/assets/stylesheets/petitions/views/_home.scss
+++ b/app/assets/stylesheets/petitions/views/_home.scss
@@ -59,7 +59,7 @@
     margin: 0;
   }
 }
-.threshold-panel:hover, .threshold-panel:focus {
+a.threshold-panel:hover, a.threshold-panel:focus {
   background-color: darken($petitions-panel-colour, 5%);
 }
 .threshold-panel-response {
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/petitions/views/_home.scss
```diff
@@ -87,7 +91,15 @@
   }
 }
 
-.local-to-you h2,
-.start-petition h2 {
-  @include bold-24;
+#local-to-you-heading,
```
alexander-bobin: We should not use the #anchors for headings. Better to use the classes as was previously done.

alexander-bobin:
Generally LGTM. A shame we don't have any integration tests for it though.

app/assets/stylesheets/petitions/views/_home.scss
henryhadlow: Oops! Didn't mean to commit that. Will change that back.

henryhadlow:
Just updated to address your previous line note.

Not that good at tests. We can revert this shortly after launch if you like.

alexander-bobin:
OK. I'll merge on the understanding that the code looks good.

PR #296, opened 2015-07-09
Remove resend confirmation email

pixeltrix:
This feature could be exploited by someone to spam people about their petition and was confusing to users. If someone hasn't received their email they can simply re-sign the petition.

https://www.pivotaltracker.com/story/show/98728106


yahoopete:
I think we can get rid of e-petitions/app/views/petition_mailer/no_signature_for_petition.html.erb: and e-petitions/app/views/petition_mailer/no_signature_for_petition.text.erb too?

PR #301, opened 2015-07-10
Bug snagging

pixeltrix:
- [x] UK citizen validation on sign skews page:  
      https://www.pivotaltracker.com/story/show/98640970
- [x] Sponsor message on email confirm:  
      https://www.pivotaltracker.com/story/show/98637502
- [x] Sponsor limit:  
      https://www.pivotaltracker.com/story/show/98647724
- [x] 6th (or more) sponsor to sign after moderation:  
      https://www.pivotaltracker.com/story/show/98638702
- [x] Email to creator when sponsor signs an open petition:  
      https://www.pivotaltracker.com/story/show/98794612

Don't merge until all boxes are ticked

### DIFF
--- a/app/assets/stylesheets/petitions/_forms.scss
+++ b/app/assets/stylesheets/petitions/_forms.scss
@@ -31,6 +31,8 @@ select.form-control:hover {
   @extend %outdent-to-full-width;
 
   @include media(tablet) {
+    margin-left: -$gutter;
+    margin-right: -$gutter;
     padding-left: $gutter - 5px;
     padding-right: $gutter;
   }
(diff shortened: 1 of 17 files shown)


app/assets/stylesheets/petitions/_forms.scss
```diff
@@ -31,11 +31,18 @@ select.form-control:hover {
...
 
   @include media(tablet) {
+    margin-left: -$gutter;
+    margin-right: -$gutter;
     padding-left: $gutter - 5px;
     padding-right: $gutter;
   }
 }
 
+.form-group + input[type=submit] {
```
alexander-bobin: The problem is caused by the fact that `.form-group` is floated left in gov uk elements. Instead of removing the float from the `.form-group` (I suspect that's there for a reason) we could add a `clear: both;`. This seems to do the trick.

pixeltrix:
Updated to address @alexander-bobin's comment

PR #311, opened 2015-07-13
Add Contact page, remove Feedback function

henryhadlow:
The Petitions Committee requested we publish their email and phone number.
This negates the need for a feedback function.
If they decide they don't it, we can always `git revert`


db/migrate/20150713143149_remove_feedback_email_from_site.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveFeedbackEmailFromSite < ActiveRecord::Migration
```
alanth: As there is a possibility we'll need to revert this, should we have an up/down on migrations? @pixeltrix

yahoopete:
I don't think this negates the need for feedback.  I'd just give people options,  do. Some may like to leave anonymous feedback,  this route helps that too. 

I'd put this on ice.

henryhadlow:
OK. There was another possible treatment which I'll put in a separate PR and you can have a look.

henryhadlow:
Replaced this with hub pull-request
https://github.com/alphagov/e-petitions/pull/315

PR #314, opened 2015-07-14
Update analytics code

alanth:


yahoopete:
This changes the cookies that are set,  so we will need to update that page.... 

Also @pixeltrix should look at this,  he  changed something here recently,  might be worth checking what that was...

pixeltrix:
@pete I just moved the file from layouts so that the Action View automatic template digest function could find the template.

@alanth can you update the help page and I will merge

yahoopete:
Do we have this script on the error pages?  Be useful to gauge those
too....

PR #315, opened 2015-07-14
Add contact info to feedback page

henryhadlow:
Replaces https://github.com/alphagov/e-petitions/pull/311
Also: make buttons not full-width at a narrower viewport size


henryhadlow:
Replaces https://github.com/alphagov/e-petitions/pull/311

PR #317, opened 2015-07-14
Make preparations

pixeltrix:
Some things that we'll need for production when we go live


oskarpearson:
All looks good to me.

PR #322, opened 2015-07-15
Add home_page? method to error view context

pixeltrix:
THe header partial now uses the home_page? method to decide whether to make the site name a link or not so we need to add that to our context.

https://www.pivotaltracker.com/story/show/99040670


alanth:
:+1: will merge when travis does it's thing

PR #328, opened 2015-07-18
Fix bug where video didn't scale properly for mobile

henryhadlow:
Also: Fix bug where animated numbers might display funny in older browsers


yahoopete:
Cool.

PR #329, opened 2015-07-20
Fix incorrect caching of social meta tags

pixeltrix:
The change in e7167a8 made the content of the :head fragment dependent on the page title and whether the page is an archived petition or not. However these aren't part of the cache key for that fragment so we need to add them to the list of dependencies.

https://www.pivotaltracker.com/story/show/99397764


oskarpearson:
I'm not overly familiar with how the caching key stuff works - but it looks good to me.

PR #333, opened 2015-07-20
Make words wrap nicely over the line

henryhadlow:
* Remove automated approach using buddysystem.js – it broke for single words marked up with both an a and an h1/h2/h3 tag
* Add non-breaking spaces using \xA0 character
* Turn off hyphenation


yahoopete:
LGTM!

henryhadlow:
"Let's get this merged"

PR #335, opened 2015-07-20
Move browserconfig.xml and manifest.json to actions

pixeltrix:
The application manifest files need to be generated with asset urls so it's simplest to just make them into Rails actions.


oskarpearson:
Looks ok to me

PR #337, opened 2015-07-21
Add error pages for 400 and 406 errors

pixeltrix:
Also add a redirect to handle the most common form of 406 error.

https://www.pivotaltracker.com/story/show/99489826


alanth:
LGTM

PR #338, opened 2015-07-21
Ph homepage thresholds

yahoopete:
Updated to reference the fact that there are no responses or debates, not that there are no petitions with that sig count.


alanth:
LGTM

PR #340, opened 2015-07-21
Minor grammar tidy-up

alanth:


yahoopete:
I think how it was is legit.... it came from our content designer.... :)

Totally fine to start sentence with But.

http://www.getitwriteonline.com/archive/032601startsentandbut.htm

alanth:
I'll leave it

PR #342, opened 2015-07-22
Follow "Tell, don't ask" principle for unsubscribe

pixeltrix:
IT is not the job of the controller to inquire about the state of the signature and act appropriately - the logic should be in the model.

https://pragprog.com/articles/tell-dont-ask


alanth:
LGTM

PR #344, opened 2015-07-22
Ignore 400 Bad Request type errors

pixeltrix:
There are some situations where a request is bad that Rails will blow up with an exception. The public exceptions middleware handles these correctly and returns a 400 Bad Request page but they still get reported by default to AppSignal so we should probably ignore them since we'll probably ignore any emails that get sent for those anyway.


oskarpearson:
AOK with me

PR #350, opened 2015-07-22
Fix incorrect class name in ignore_exceptions config

pixeltrix:
The correct name for the class is ActionController::BadRequest and not ActionDispatch::BadRequest.


alanth:
:+1:

PR #351, opened 2015-07-22
Rescue Faraday::ClientError as well as Faraday::ResourceNotFound

pixeltrix:
If someone enters garbage into the postcode field it can cause the parliament api to return a 400 Bad Request which Faraday raises as a ClientError. We don't want to know about these so rescue them along with the 404 Not Found responses.


alanth:
Nice.

PR #352, opened 2015-07-23
Lock signature row before verifying

pixeltrix:
There exists a race condition where a verification link is used twice at the same time can result in a double signature count increment. To fix this we need to obtain a row level lock on the signature record and then check to see if the signature is pending.

This change shouldn't affect performance significantly as concurrent access to the signature record will only occur when the background job that validates signature counts runs which is every 30 minutes or the verification link is used twice which is what we want to prevent.

The tests needed to be tweaked because `with_lock` reloads the record from the database so in memory changes will have been lost.

### DIFF
--- a/app/models/petition.rb
+++ b/app/models/petition.rb
@@ -362,10 +362,6 @@ class Petition < ActiveRecord::Base
     creator_signature && creator_signature.validate! && reload
   end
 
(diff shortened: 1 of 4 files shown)


oskarpearson:
This looks fine to me. I've done some testing at the SQL level and it looks good.

PR #353, opened 2015-07-23
Content changes

henryhadlow:
* Add government response summary to list of petitions with a government response
* Add wording 'Waiting *for* 6 days for a...' because people read it as a wait time
* Treat different petition states differently on the 'All petitions' list
* Add 'All petitions run for 6 months' message to the deadline on open petition show
* Apply `auto_link` and `simple_format` to petition rejection text

### DIFF
--- a/app/assets/stylesheets/petitions/views/_petition-show.scss
+++ b/app/assets/stylesheets/petitions/views/_petition-show.scss
@@ -51,9 +51,13 @@
     @include bold-24;
     margin-bottom: $gutter-two-thirds;
   }
-  .label {
+  .label, .note {
     display: block;
     @include core-16;
     margin-bottom: 5px;
+  }
+  .note {
+    color: $secondary-text-colour;
+    margin-top: 10px;
   }
 }
--- a/app/views/admin/petitions/_petition_details.html.erb
+++ b/app/views/admin/petitions/_petition_details.html.erb
@@ -2,14 +2,6 @@
   <dt>Status</dt>
   <dd class="petition-meta-state"><%= @petition.state.capitalize %></dd>
 
-<% if @petition.rejected? or @petition.hidden? -%>
-  <dt>Rejection reason</dt>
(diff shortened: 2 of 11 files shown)


pixeltrix:
@henryhadlow for some reason there's cucumber scenario that checks that a rejection reason has HTML tags escaped and does not autolink URLs. Since those reasons are entered by moderators this seems somewhat crazy since we autolink stuff from the general public. If we want moderators to be able to enter HTML you can pass a `sanitize: false` option to `simple_format` and you'll need to change the cucumber scenario to match our new expectations.

The other failure is because you've removed the signature count from responded petitions - was this intentional? If so then you'll need to change the expectation.

henryhadlow:
Thanks @pixeltrix. I'll try and look at this tomorrow morning but if not
I'll pick it up on Tuesday.

I'd rather just apply simple_format and auto_link to the rejection reason
(also – won't the raw get sent to petition creators?)

I thought the number of signatures was less relevant than the response date
and summary so I changed it. I'll try and change that test too, but I don't
know where these fox petitions are coming from...?

henryhadlow:
Actually I'm not sure how to fix this :(

How are the test petitions created? If I knew that then I could figure out what the test should expect...

pixeltrix:
@henryhadlow it's in the scenario itself - changing it to this works:

``` cucumber
Scenario: Suzie sees reason for rejection if appropriate
  Given a petition "Please bring back Eldorado" has been rejected with the reason "We like http://www.google.com and bambi@gmail.com"
  When I view the petition
  Then I should see the petition details
  And I should see the reason for rejection
  And I should see "We like http://www.google.com and bambi@gmail.com"
  And I should see a link called "http://www.google.com" linking to "http://www.google.com"
  And I should see a link called "bambi@gmail.com" linking to "mailto:bambi@gmail.com"
  And I should not see the vote count
  And I should see submitted date
  And I cannot sign the petition
```

henryhadlow:
Thanks @pixeltrix

PR #354, opened 2015-07-23
Fix page title on local petitions without a postcode

pixeltrix:
If the user doesn't enter a postcode when searching from the home page or they navigate to the page directly then the page title helper doesn't have a count to pick which translation to use. For this case set the count to -1 so that the :other key is used from the translation.


pixeltrix:
This is what you currently get on production:

bad-title

alanth:
LGTM

PR #356, opened 2015-07-24
Increase length of the government response field

henryhadlow:
* Government response summary now 200 characters
* Government response details now 5000 characters


alanth:
:+1:

PR #358, opened 2015-07-25
Made the requested time suppliable by user

oskarpearson:
Made the requested time suppliable by user

If we need to retry the job due to technical problems, we need to be able to force the timestamp.


pixeltrix:
LGTM :shipit:

PR #360, opened 2015-07-28
Refactor the jobs that send email.

ansonK:
Break the jobs that deliver email updates to petition signatories into 2 parts:

* a parent job that will load the recipients and enqueue a child job for each
* a child job that will perform a single email delivery

Doing this removes the memory leak seen in previous version when delivering emails, and gives greater control over how much work is left to do as each job now does 1 thing and will eliminate the need for long running mail delivery jobs.

This PR should be merged on top of #359 

Multiple worker processes can spread the load of the actual email sending, rather than just tying up 1 process for a long time.

Both the parent and child jobs will check that the petition they are delivering information for and will exit if the petition has been updated since the jobs were enqueued.

Tried to keep with the existing structure, could do with further refactoring.

Tested locally delivering 10K emails:

* Worker Memory levels at ~250MB when enqueuing mailing jobs (previously was unbounded growth)
* Worker Memory evens out at ~270MB when running mailing jobs
* Mailing jobs run at ~7 - ~10 job/s (with local mailhost)

Update:
Deliver email jobs are now idempotent. They can be run multiple times but only send an email once (until the petition is updated)

Things this PR does not address:

* there is no check to confirm that if a petition has X signatories, X emails actually got sent.

/cc @h-lame @oskarpearson @alanth @pixeltrix


app/jobs/email_delivery_jobs/base.rb
```diff
@@ -0,0 +1,69 @@
...
+      requested_at_as_string:, mailer: PetitionMailer.name, logger: nil)
+
+      @mailer = mailer.constantize
+      @signature = signature
+      @petition = petition
+      @requested_at = requested_at_as_string.in_time_zone
+      @timestamp_name = timestamp_name
+      @logger = logger
+
+      if petition_has_not_been_updated?
```
h-lame: I think we can make this job idempotent if we made it also check the receipt timestamp for the signature and return if the stamp is not `petition_timestamp`.  The enqueue job already filters them out via the `petition.signatures_to_email_for` call, but we can do a check here too.

h-lame:
Are we sure this eliminates the memory leak?  @pixeltrix narrowed it down to the double enqueue, which we're still doing and this branch doesn't sit on top of his one (see #359).  Not that we should avoid merging as it has merits above fixing the memory leak, but we should check it with more signatures first I guess?

ansonK:
> Are we sure this eliminates the memory leak? @pixeltrix narrowed it down to the double enqueue, which we're still doing and this branch doesn't sit on top of his one (see #359). Not that we should avoid merging as it has merits above fixing the memory leak, but we should check it with more signatures first I guess?

Absolutely should test with more emails. Memory held steady across 10K but may grow with more than that. Should use @pixeltrix's fix anyway

config/application.rb
```diff
@@ -44,5 +44,8 @@ class Application < Rails::Application
     )
 
     config.action_dispatch.default_headers.merge!('X-UA-Compatible' => 'IE=edge')
+
+    # Load the job concerns
+    config.eager_load_paths += %W(#{config.root}/app/jobs/**/*)
```
pixeltrix: Don't think this is needed - everything under `app` should be eager loaded

config/initializers/job_logger.rb
```diff
@@ -0,0 +1,5 @@
+class ActiveJob::Base
+  def logger
+    @logger || JobLogger.new(self.class.name)
+  end
```
pixeltrix: Are you intending to override all of the Active Job logging here?

spec/support/active_job.rb
```diff
@@ -1,3 +1,8 @@
 RSpec.configure do |config|
   config.include(ActiveJob::TestHelper)
+
+  config.before(:each) do
+    ActiveJob::Base.queue_adapter.enqueued_jobs = []
+    ActiveJob::Base.queue_adapter.performed_jobs = []
+  end
```
pixeltrix: The `ActiveJob::TestHelper` has a [`before_setup`][1] method that clears both of these. If you're thinking but that's Minitest and not RSpec then don't worry - RSpec Rails calls it [here][2]. :smile:

[1]: https://github.com/rails/rails/blob/master/activejob/lib/active_job/test_helper.rb#L22-L23
[2]: https://github.com/rspec/rspec-rails/blob/master/lib/rspec/rails/adapters.rb#L126

spec/support/email.rb
```diff
@@ -1,4 +1,8 @@
 RSpec.configure do |config|
   config.include(EmailSpec::Helpers)
   config.include(EmailSpec::Matchers)
+
+  config.before(:each) do
+    ActionMailer::Base.deliveries.clear
+  end
```
pixeltrix: Maybe we should include [`EmailSpec::MailerDeliveries`][1] and use that instead of dealing with `ActionMailer::Base.deliveries`, e.g:

```
config.before(:each) do
  reset_mailer
end
```

and

```
expect { perform_job }.to change(all_emails, :size).by(1)
```



[1]: https://github.com/bmabey/email-spec/blob/master/lib/email_spec/deliveries.rb

app/jobs/email_delivery_jobs/base.rb
```diff
@@ -0,0 +1,82 @@
...
+    attr_reader :mailer, :signature, :timestamp_name, :petition, :requested_at
+
+    def can_send_email?
+      petition_has_not_been_updated? && email_not_previously_sent?
+    end
+
+    def send_email
+      create_email.deliver_now
+
+      rescue Errno::ECONNREFUSED, Errno::ETIMEDOUT, Net::SMTPError, Timeout::Error => e
```
pixeltrix: Have you looked at the AppSignal list of errors we've received - we may want to adjust our handling based on what error we receive.

app/jobs/email_petition_signatories_job.rb
```diff
@@ -0,0 +1,80 @@
+class EmailPetitionSignatoriesJob < ActiveJob::Base
+  queue_as :default
+
+  def self.run_later_tonight(petition, requested_at, *extra_args)
+    set(wait_until: later_tonight).
+      perform_later(petition, requested_at.getutc.iso8601, *extra_args)
+  end
```
pixeltrix: We don't need to override this in all of the subclasses since we define a method `timestamp_name` anyway - also the `*extra_args` is never used as far as I can see. e.g.

``` ruby
def self.run_later_tonight(petition, requested_at = Time.current)
  petition.set_email_requested_at_for(timestamp_name, to: requested_at)
  perform_later(petition, requested_at.getutc.iso8601(6))
end
```

Note that I've changed the `iso8601` call to include a usec timestamp so we don't need a `to_i` comparison later

app/jobs/email_petition_signatories_job.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  # admins can ask to send the email multiple times and each time they
+  # ask we enqueues a new job to send out emails with a new timestamp
+  # we want to execute only the latest job enqueued
+  def petition_has_been_updated?
+    # NOTE: to_i comparison is used to cater for precision differences
+    # between DB timestamp (petition_timestamp precise to fractional
+    # seconds) and job timestamp (requested_at - only seconds precise)
+    petition_timestamp.to_i != requested_at.to_i
```
pixeltrix: If you make the change in my comment above to use `iso8601(6)` you can change this to:

```
def petition_has_been_updated?
  (petition_timestamp - requested_at).abs > 1
end
```

spec/support/email.rb
ansonK: Started doing this but quickly realised its a far bigger job - most controller specs make assertions about emails sent and none of them use the email_spec helpers.

Refactoring all of them would be a PR in itself

ansonK:
@pixeltrix I've made the requested changes (except using email_spec helpers as that is a larger chunk of work)

Also refactored out using base classes for the email sending jobs in favour of concerns. Mainly due to the fact that the base classes could be be instantiated themselves (and would not make sense to)

I've left the new commits unsquashed to make it easier to see what I've changed, once everyones happy then I'll squash

ansonK:
> I seem to remember trying iso8601(n) for comparisons somewhere, and having to not using because of rounding. Something like ruby doesn't round so a usec of 1178 passed to iso8601(2) is 11 not 12 but I think the times coming from PG do round? I can't really remember the exact details, but reading this I do remember trying it and having problems.

There are specs that check that the timestamps do match (and not match) when expected:
https://github.com/ansonK/e-petitions/blob/email-send-refactor/spec/jobs/shared_examples.rb

Done a quick console test and the rounding seems to work the same way .. ?:
```
irb(main):023:0> time = Time.current
=> Wed, 29 Jul 2015 13:28:37 BST +01:00
irb(main):024:0> Petition.last.update! created_at: time
  Petition Load (0.7ms)  SELECT  "petitions".* FROM "petitions"  ORDER BY "petitions"."id" DESC LIMIT 1
   (0.2ms)  BEGIN
  SQL (1.0ms)  UPDATE "petitions" SET "created_at" = $1, "updated_at" = $2 WHERE "petitions"."id" = $3  [["created_at", "2015-07-29 12:28:37.916448"], ["updated_at", "2015-07-29 12:28:39.594354"], ["id", 100000]]
   (0.8ms)  COMMIT
=> true
irb(main):025:0> petition = Petition.find 100000
  Petition Load (0.5ms)  SELECT  "petitions".* FROM "petitions" WHERE "petitions"."id" = $1 LIMIT 1  [["id", 100000]]
=> #<Petition id: 100000, action: "Incidunt delectus velit id suscipit nostrum eos do...", additional_details: "Aliquam porro asperiores dolores praesentium. Sed ...", state: "pending", open_at: nil, creator_signature_id: 1, created_at: "2015-07-29 12:28:37", updated_at: "2015-07-29 12:28:39", closed_at: nil, signature_count: 0, notified_by_email: false, background: "Ea nobis enim architecto labore ratione ullam dolo...", sponsor_token: "K3fnmTnaq2iw8GSF4uTV", government_response_at: nil, scheduled_debate_date: nil, last_signed_at: nil, response_threshold_reached_at: nil, debate_threshold_reached_at: nil, rejected_at: nil, debate_outcome_at: nil, moderation_threshold_reached_at: nil>
irb(main):026:0> time.getutc.iso8601(6)
=> "2015-07-29T12:28:37.916448Z"
irb(main):027:0> petition.created_at.getutc.iso8601(6)
=> "2015-07-29T12:28:37.916448Z"
irb(main):028:0> time.getutc.iso8601(2)
=> "2015-07-29T12:28:37.91Z"
irb(main):029:0> petition.created_at.getutc.iso8601(2)
=> "2015-07-29T12:28:37.91Z"
```

h-lame:
Ah, yes, maybe it's because I wasn't pushing everything through an iso8601 call so there were still some discrepancies.  This is probably fine.  Forget I said anything.

PR #362, opened 2015-07-29
Make <details> open if there is an #id in the URL

henryhadlow:


app/views/pages/_home_responded_petitions.html.erb
```diff
@@ -7,7 +7,7 @@
         <h3><%= link_to petition.action, petition_path(petition), class: "threshold-petition-title" %></h3>
         <p class="intro">The government responded</p>
         <blockquote class="pull-quote"><%= simple_format(petition.government_response.summary) %></blockquote>
-        <p><%= link_to "Read the response in full", petition_path(petition) %></p>
+        <p><%= link_to "Read the response in full", petition_path(petition, show_response: "✓", anchor: 'response-threshold') %></p>
```
oskarpearson: Does this work on browsers that don't support unicode? I'm not sure of our browser coverage expectations.

oskarpearson:
I'm probably not the best person to review the browser polyfill - but other than my query above, I think this is fine :)

app/views/pages/_home_responded_petitions.html.erb
henryhadlow: Not sure. I'll remove the unicode character.

pixeltrix: @oskarpearson are there browsers that we support that don't support unicode? Even IE6 has UTF-8 support.

PR #363, opened 2015-07-30
Add links to petitions waiting for X to homepage

henryhadlow:
It's a bit misleading to only show the petitions that *have* a response or debate outcome

### DIFF
--- a/app/helpers/home_helper.rb
+++ b/app/helpers/home_helper.rb
@@ -10,6 +10,10 @@ module HomeHelper
       actioned.fetch(state, 0)
     end
 
+    def with_result
+      {:with_response => actioned[:with_response], :with_debate_outcome => actioned[:with_debate_outcome]}
+    end
+
     private
 
     def actioned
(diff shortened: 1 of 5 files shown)


app/helpers/home_helper.rb
```diff
@@ -18,9 +22,13 @@ def actioned
 
     def generate_actioned
       scope, actioned = Petition.visible, {}
+      awaiting_response = scope.awaiting_response.awaiting_response.by_waiting_for_response_longest
```
oskarpearson: I think you've got an extra awaiting_response scope here? Not sure if that was intentional.

```
scope.awaiting_response.awaiting_response.by_waiting_for_response_longest
should be
scope.awaiting_response.by_waiting_for_response_longest
```

oskarpearson:
I'm a bit concerned that we're doing a lot more SQL queries on the homepage. I think it might be worth checking the performance impact of this before it hits prod.. Is the home page cached, so that these queries only run occasionally?

There may be indexes that we need to create to make sure that the queries are against indexed fields.

The stats pages don't show a lot of hits of the homepage, so I don't know that it'd be a huge problem... but I think it's worth testing.

henryhadlow:
Over the last week, homepage got 150k pageviews, 3.5% of traffic

pixeltrix:
@oskarpearson @henryhadlow the home page is cached with a five minute expiry and is dependent on the `last_signed_at` column of the `Site` model so the answer is yes, but how effective that caching is could be debated since there are always new signatures happening.

henryhadlow:
@pixeltrix thanks :) can we merge it?

pixeltrix:
@henryhadlow you haven't fixed the double `awaiting_response` scope yet

henryhadlow:
ohhhhh yeahhhhhhh. sorry.

PR #364, opened 2015-07-30
Redirect to clean petition url

ansonK:
Adds a redirect back to the petition show page for users that reuse the verify link in the sign petition confirmation email.

If the signer shares the link to others then everyone else will also be redirected to the petition show page

https://www.pivotaltracker.com/story/show/100119558


h-lame:
:+1:

PR #365, opened 2015-07-30
Improve reliability of AppSignal initialization process

pixeltrix:
Since Puma detects running threads when preloading an application we need to stop it and then restart when each of the workers are forked. It was sort of working before because it detected when the PID of the process had changed and then restarted the thread itself.


oskarpearson:
Looks good to me.

PR #366, opened 2015-07-31
Add Committee message to email about government response

henryhadlow:


henryhadlow:
For some reason I can't use `petition_mailer_preview` to test this. Can someone take a look?

pixeltrix:
@henryhadlow you need to add a method that generates the preview email to `PetitionMailerPreview`, e.g:

```
  def notify_signer_of_threshold_response
    petition = Petition.with_response.last
    signature = petition.signatures.validated.last

    PetitionMailer.notify_signer_of_threshold_response(petition, signature)
  end
```

Obviously you need to have a petition with a response in your local database.

app/views/petition_mailer/notify_signer_of_threshold_response.text.erb
```diff
@@ -10,7 +10,15 @@ Government responded:
 
 Click this link to view the response online:
 
-<%= petition_url(@signature.petition) %>
+<%= link_to nil, petition_url(@signature.petition, reveal_response: "yes") %>
```
pixeltrix: This is the plain text part of the email - we don't want HTML in here.

PR #369, opened 2015-08-06
Add reveal_response to cache keys

pixeltrix:
When sending out response emails we want to display the response already revealed but it may have been already cached so add a cache key to differentiate between the two views.


alanth:
Given the context of what we're trying to achieve, makes sense to me

PR #370, opened 2015-08-07
Add petition emails 

pixeltrix:
Adds the ability to send custom emails to petitioners using the same mechanism as the other bulk email tasks. Also updates the default feedback email address to the correct one.

https://www.pivotaltracker.com/story/show/98891944

### DIFF
--- a/app/assets/stylesheets/petitions/admin/views/_shared.scss
+++ b/app/assets/stylesheets/petitions/admin/views/_shared.scss
@@ -20,6 +20,10 @@
     margin-bottom: $gutter-half;
   }
 
+  .character-count {
+    margin-bottom: 0;
+  }
+
   // Pagination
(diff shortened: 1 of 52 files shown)


henryhadlow:
LGTM :+1:

oskarpearson:
A bit of a "meta" comment - you can ignore if you don't agree! :)

Currently there's a bunch of logic in the code to discard all but the last queued job for sending emails - effectively trying to make the jobs idempotent.

By putting only the petition ID into the arguments, it reduced the data encoded into the job. That meant that the job needed to get the latest information from the DB, which meant it always had the latest data at runtime.

My only concern with this is that if somehow we update the data and don't re-queue the job, or the logic in petition_has_been_updated is buggy - which would lead to an out of date email being sent.

PR #371, opened 2015-08-10
JSON API for petitions + CSV petitions download 

ansonK:
public petitions list with JSON
public petitions detail with JSON
download petitions list from admin as CSV

Implements:

* https://www.pivotaltracker.com/story/show/100635802
* https://www.pivotaltracker.com/story/show/100635838
* https://www.pivotaltracker.com/story/show/100547702

/cc @pixeltrix

PR #372, opened 2015-08-12
Streaming csv file downloads

ansonK:
Steams the admin CSV download so large numbers of Petitions can be downloaded as CSV without performance penalties.

Loads petitions in batches of 1000 from the database and streams each petition in CSV format to the client

Respects the same search queries as the UI (state,keywords etc)
Skips the pagination of Petitions (Browseable::Search now has a #in_batches method)

Also changes the PetitionCSVPresenter to use CSV::Row internally to handle csv formatting and escaping.

Note: #371 should be merged first. This PR is based off that

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -336,4 +336,4 @@ DEPENDENCIES
   will_paginate
 
 BUNDLED WITH
-   1.10.4
+   1.10.6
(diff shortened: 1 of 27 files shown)


pixeltrix:
CSV streaming stuff seems :+1:

alanth:
Testing the branch on preview also looks good. I'm going to merge

PR #377, opened 2015-08-26
Splitting Israel and The Occupied Palestinian Territories' into their…

yahoopete:
… own country options.


alanth:
Merging. Caveat is that old data with the "old country" will remain as is

PR #379, opened 2015-09-03
Update use of 'government' based on the GDS style guide

maxehmookau:
On the home page, the phrase `petitions got a response from the Government` refers to the government with an upper-case G. 

The style guide (https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style) says that government should _always_ be referred to with a lower-case g even when referring to a specific government. This PR simply replaces them.

### DIFF
--- a/config/locales/petitions.en-GB.yml
+++ b/config/locales/petitions.en-GB.yml
@@ -131,12 +131,12 @@ en-GB:
     counts:
       awaiting_response:
         html:
-          one: "<span class=\"count\">%{formatted_count}</span> petition is waiting for a response from the Government"
-          other: "<span class=\"count\">%{formatted_count}</span> petitions are waiting for responses from the Government"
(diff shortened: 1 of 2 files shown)


pixeltrix:
There are a few other places where we use a capital G - specifically the help page and emails. I think @yahoopete and @henryhadlow are the best people to weigh in on whether we want to do this globally.

maxehmookau:
Okey doke. Obviously not a big deal, just noticed a slight inconsistency.

henryhadlow:
We're going to go with Government when referring to the UK Government. This is because we frequently refer to Government and Parliament together and it looks weird if they're inconsistent.
This service isn't part of GOV.UK therefore isn't bound to follow the GDS style guide.

maxehmookau:
Fair enough, that's fine :)

However, if it's not part of gov.uk, what is it part of?

PR #382, opened 2015-09-04
Set CORS headers for json get requests

charlieegan3:
Make the data available from browsers via frontend apps.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -9,12 +9,18 @@ class ApplicationController < ActionController::Base
   before_action :set_seen_cookie_message, if: :show_cookie_message?
   helper_method :show_cookie_message?, :public_petition_facets
 
+  before_action :set_cors_headers, if: :json_request?
+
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -51,6 +57,12 @@ def set_seen_cookie_message
     cookies[:seen_cookie_message] = { value: 'yes', expires: 1.year.from_now, httponly: true }
   end
 
+  def set_cors_headers
+    headers['Access-Control-Allow-Origin'] = '*'
```
ansonK: Are we sure we want to allow any domain here? There isn't going to be any performance impact by opening the API to be loaded on any domain?

pixeltrix: Possibly - but it's open data which means everyone has access, plus they could just proxy requests through their own server anyway.

alanth:
Restricts to get requests and is an established pattern. I can't see any harm in this

PR #383, opened 2015-09-08
Add the ability to delete a signature from the admin

alanth:
An increasing number of people want to remove their signature from a
petition so to make life easier add a feature to the search results
page in the admin to remove a signature. Note that petition creators
can't have their signature removed.

https://www.pivotaltracker.com/n/projects/307301/stories/102570760


alanth:
LGTM

PR #384, opened 2015-09-08
Disable Appsignal monitoring for bulk email sending

pixeltrix:
When sending large email responses (greater than 200,000 emails) the job appears to lock up at the end of queuing all the emails. Since we are also seeing a great deal of memory usage by the delayed_job processs our working theory is that Appsignal is tracking all of the SQL queries and then trying to send gigabytes of data to the API causing it to hang.

We are attempting to workaround this issue by suspending instrumentation for the period of the loop - hopefully this will still give us some usable information such as the job duration and queue time.

Our initial source for this information was appsignal/appsignal#57


alanth:
LGTM

oskarpearson:
LGTM too

PR #385, opened 2015-09-08
Adjust delayed_job config to something more sane

pixeltrix:
The configured max runtime and max attempts were left as set in the initial code dump fron the first version of the website. The rationale for the chosen figures is lost in the mists of time but seem somewhat large.

The 1 week max runtime may have been to send out the bulk emails which is understandable but will be not tolerable for users. The 100 max attempts coupled with the backoff inherent in delayed_job means that the job would be in the queue for over 61 years! Perhaps earlier versions of delayed_job didn't have such a backoff? The figure of 15 for max attempts will mean a job is re-tried for over 35 hours which is a reasonable period of time.


alanth:
Looks fine as discussed.merging.

PR #386, opened 2015-09-08
Fix Appsignal when used with multiple DJ workers

pixeltrix:
Similar to the fix for Puma we need to let Appsignal know that the process has been forked so that it can re-establish connections. Unfortunately there is no good way to do this so we have to override the after_fork class method in an initializer to call Active Record's establish_connection as well as the Appsignal forked! method.

For some strange reason I was under the impression that the `bin/delayed_job` script was called 4 times when I wrote the previous fix in 6caec23 but on reflection that make no sense since it couldn't take advantage of copy-on-write memory saving.


alanth:
LGTM

PR #387, opened 2015-09-10
The reveal_response is a cache key for petition pages

pixeltrix:
The key was incorrectly used on the home page when it should be the petition page


yahoopete:
Is this for the social sharing thing, or the linking to the detailed response thing? Or both? :)

alanth:
@yahoopete This is for detailed response thing

alanth:
LGTM

PR #388, opened 2015-09-11
[WIP] Sidekiq

ansonK:
First pass at replacing Delayed job with Sidekiq / RFC

/cc @pixeltrix @oskarpearson @alanth 

* sidekiq web ui available at /admin/sidekiq
* sidekiq web ui only visible to logged in admin users with sysadmin role
* each job is enqueued to its own queue for better visibility and for particular job types to be prioritized
* default job retry set to 15
* signature confirmation emails enqueued into their own queue (rather than generic 'mailer' queue)

Requires a ```REDIS_URL``` ENV variable as detailed here: https://github.com/mperham/sidekiq/wiki/Using-Redis#using-an-env-variable

Run worker process with:
```$ sidekiq``` to run all queues
```$ sidekiq -q queue_name -q other_queue_name``` to process only the named queues

Still todo:
* group job types into queues - high, med, low queues rather than each job type in its own queue?
* init scripts for sidekiq (esp queue priority, concurrency etc)
* redis infrastructure (multi a-z deployment?)

## Benchmarks:
1. Enqueue 100K deliver_email type jobs
2. Deliver 100K deliver_email type jobs

### Setup:
* 1 Petition, 100K signatures
* 8 core MacPro 16GB RAM
* local postgres db
* local SMTP 'black hole' 
* local redis server

### Delayed Job:
Test 1: Enqueue 100K deliver_email type jobs
```Performed EmailDebateScheduledJob from DelayedJob(default) in 363246.3ms (100 minutes)```

Test 2: Deliver 100K deliver_email type jobs
(4 Delayed jobs processes)
```
real	29m46.365s (this does not look right. Needs re-running as must be > 50min)
user	50m15.048s
sys     5m19.807s
```
Memory was constant at ~260MB per process

### Sidekiq:
Test 1: Enqueue 100K deliver_email type jobs
```EmailDebateScheduledJob JID-420bc80200ef6c4ccf06f603 INFO: done: 88.376 sec```

Test 2: Deliver 100K deliver_email type jobs
(4 Sidekiq processes, 20 threads each)
```
real       16m42.842s
user       0m3.086s
sys        0m0.940s
```
Memory was constant at ~275MB per process
All 4 processes were CPU bound (100%+) and SMTP black hole was also 100% so the test was bound by those factors

### CODE REVIEW

**Comment by h-lame** [comment]
Certainly compelling on the performance front.  We could run all this on the same infrastructure by installing redis on the worker box, or do we look at some kind of AWS hosted redis to we get some kind of fault tolerance.

Re: the new job would doing this: 
```ruby
def send_email_to_petition_signer(signature)
  PetitionMailer.email_confirmation_for_signer(signature).deliver_later(queue: :deliver_confirmation_email)
end
```
be of any benefit?  So it doesn't stick out as "odd" when no other emails are queued via their own job.  We've already got the abstraction of `send_email_to_petition_signer` to hide how we send the email anyway.

ansonK:
> Re: the new job would doing this:
> 
> def send_email_to_petition_signer(signature)
>   PetitionMailer.email_confirmation_for_signer(signature).deliver_later(queue: :deliver_confirmation_email)
> end
> be of any benefit?

Actually went that way first but then made the change so that all emails are wrapped in a deliver_* type job to be consistent. Also thought that exposing the queue like that in the controller was unneeded. Happy to change back if prevailing opinion swings that way?

oskarpearson:
I'm definitely up for switching to Sidekiq.

We can't run the redis cache on the box itself - the worker boxes are fungible, and the queued jobs are not. If an availability zone goes down we shouldn't lose all queued jobs. Also: the webservers would need to service-discover the worker boxes so that they could queue jobs on them.

We'd almost certainly use AWS elasticache with Redis https://aws.amazon.com/blogs/aws/elasticache-redis-multi-az/ and integrate it into the CloudFormation config ala http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticache-cache-cluster.html

oskarpearson:
BTW: I looked at the Redis replication stuff around go-live time as an alternative to Memcache, and hit this: https://forums.aws.amazon.com/thread.jspa?messageID=620524 

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticache-replicationgroup.html might be the solution, or we might need to use scripts instead of CloudFormation ala 
http://www.autoscaling.io/blog/2015/1/24/setting-up-a-clustered-redis-in-multiple-azs-in-a-vpc-with-replication

It'll take a bit of investigation

h-lame:
Actually, yeah, I think consistency wins - if all our emails are now going through a custom job with `deliver_now`, rather than `deliver_later` in a controller - that seems better to me.  You're right about exposing the queue name - that feels less than good.

pixeltrix:
Closing because I'm not convinced that we need to do this yet - #513 makes queueing the emails much quicker and we can just scale the number of workers easily now that we don't have to worry about cron jobs after #511.

PR #389, opened 2015-09-13
Finesse pages that are indexed by search engines

pixeltrix:
We don't want admin pages and new signature pages appearing in search engine results so be more explicit about what is to be indexed and what isn't.


alanth:
LGTM

PR #390, opened 2015-09-13
Add user agent string to feedback message

pixeltrix:
Feedback messages are often used to report bugs which tend to be browser dependent so rather than having to go back and ask what their browser was we can just grab the `User-Agent` header and add that to then message.

Rather than trying to parse the data in the application it's left to an administrator to use one of the many online parsers since they are more likely to be up-to-date.

One such parser can be found at: https://udger.com/resources/online-parser


alanth:
Works locally. Merging.

PR #395, opened 2015-09-21
Homepage links to government responses

henryhadlow:
On the homepage, in the 10,000 signatures section, add a query param and id to petition titles so they link to the government response directly


pixeltrix:
LGTM :shipit:

PR #396, opened 2015-09-22
Fix robots.txt

pixeltrix:
Unfortunately trying to comply with the robots.txt standard to the letter runs afoul of two problems, the first being that Google places a 500KB limit on the file and secondly Google seems to be thinking that the show page is also excluded from the index as well.

Switch back to using a static robots.txt and use the Google extension for wildcard matching URLs - we'll have to live with any other search engines that don't support wildcards indexing pages that we don't want them to.


alanth:
Simple it is then

PR #397, opened 2015-09-23
Fix caching of social meta tags

pixeltrix:
Just use the url and the updated_at timestamp of the site to generate the cache key for the head elements. Whilst this will mean that we add cached head elements for pages that will be only viewed once such as the signature validation page the LRU nature of Memcached means they will be evicted first.

Also if the @petition variable is set then use it for setting the meta tags so that even if someone shares a thank you page or a signature validation page then the URL shared by Facebook will be the main petition url and not the url of the page they last visited.


ansonK:
:+1:

PR #402, opened 2015-10-20
Personalize notifications for creators

pixeltrix:
It's impersonal to refer to creators as signers when sending notifications and it's easy to detect so create personalized emails for them.


alanth:
LGTM

alanth:
LGTM

PR #403, opened 2015-10-20
Add an admin interface for validating signatures

pixeltrix:
The moderators get requests to validate people's signatures via the feedback form for whatever reason and it's easier to validate them manually to explain that they can just re-sign and it will send them a new validation email.

https://www.pivotaltracker.com/story/show/105664320


alanth:
LGTM

PR #404, opened 2015-10-20
Prevent the use of disposable email addresses for signing

pixeltrix:
Whilst disposable email addresses do have valid uses for privacy in the case of petition signing they can be abused for bulk signing of emails via scripts. Add a simple validation check to prevent their use based on a list of domains from [martenson/disposable-email-domains][1].

[1]: https://github.com/martenson/disposable-email-domains


alanth:
LGTM

PR #405, opened 2015-10-27
Only show petitions with debate outcome on home page

pixeltrix:
There was an edge condition around petitions that had been debated but had not yet had there debate outcome added when viewing the home page and the debated list. The code assumed that there was a debate outcome present so to fix this the affected petitions will not appear on the home page until they have a positive debate outcome and the debated list will fall back to the scheduled debate date if there is no debate outcome.


alanth:
LGTM

PR #406, opened 2015-10-28
Correct section DOM id

pixeltrix:
The link at the top of the home page to the debated petitions section is tied to the scope name so we need to change the DOM id to match.

https://www.pivotaltracker.com/story/show/106483746


alanth:
Makes sense to me

PR #414, opened 2015-12-09
Move signature count updates outside the lock

pixeltrix:
Investigation shows a massive amount of locks when trying to update the signature count on the petition so move it outside of the lock but retain the pending check so that we don't count double clicks.


alanth:
:+1:

alanth:
LGTM

PR #413, opened 2015-12-09
Temporarily disable journals

pixeltrix:
Due to write contention on the AWS RDS instance temporarily disable the journals for constituency and country whilst we deal with the extreme load from https://petition.parliament.uk/petitions/114003

PR #416, opened 2015-12-11
Reinstate journalling

h-lame:
This reverts the PR to temporarily disable them[1].  We're pretty confident that moving count updates outside the lock[2] has resolved the majority of the write contention we were seeing under heavy load so we can re-instate these counts.

We also add rake tasks to regenerate the journal counts from the data: `rake epets:journals:reset_countries` and `rake epets:journals:reset_constituencies`.

[1]: https://github.com/alphagov/e-petitions/pull/413
[2]: https://github.com/alphagov/e-petitions/pull/414

### DIFF
--- a/app/models/constituency_petition_journal.rb
+++ b/app/models/constituency_petition_journal.rb
@@ -38,4 +38,19 @@ class ConstituencyPetitionJournal < ActiveRecord::Base
       raw_write_attribute(:signature_count, signature_count+1)
     end
   end
+
+  def self.reset!
+    connection.execute "TRUNCATE TABLE #{self.table_name}"
+    connection.execute <<-SQL.strip_heredoc
+      INSERT INTO #{self.table_name}
+        (petition_id, constituency_id, signature_count, created_at, updated_at)
+      SELECT
+        petition_id, constituency_id, COUNT(*) AS signature_count,
+        timezone('utc', now()), timezone('utc', now())
+      FROM signatures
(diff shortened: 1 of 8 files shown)


alanth:
Merging to put up on preview and shake down the performance

alanth:
The issue with the rake task is that if the journals are being updated, then it will fall over as the insert will fail due to duplicate key values. This form of query executes a lot faster (regardless of the size of target table) and is safe to run whilst the journals are being updated.

INSERT INTO country_petition_journals (petition_id, country, signature_count, created_at, updated_at) SELECT petition_id, country, COUNT(*) AS signature_count, timezone('utc', now()), timezone('utc', now()) FROM signatures where state = 'validated' and concat(petition_id,country) not in (SELECT concat(petition_id,country) FROM country_petition_journals) GROUP BY petition_id, country;

alanth:
The constituency insert either needs the condition "and constituency_id is not NULL" for those outside of the UK, or the country filtered to UK.

h-lame:
Oh, good point about the constituency_id filter; I forgot about that!

For the other point the `where not in` suggests we could have 1 or more rows in the journal table with the wrong count (if a petition is signed between the `truncate` and `insert` statements then we'd have a row for that with a 1 when we want to recalculate it from all the other data.  What if we wrapped the `truncate` & `insert` in a transaction or lock?  Or would that be problematic from a perf. point of view?

PR #418, opened 2015-12-14
Make recalculating journals more robust

h-lame:
To follow up on comments on #416 which has already been merged.

### DIFF
--- a/app/models/country_petition_journal.rb
+++ b/app/models/country_petition_journal.rb
@@ -25,6 +25,8 @@ class CountryPetitionJournal < ActiveRecord::Base
     end
 
     def reset!
+      # NOTE: country <> '' is the closest we can (performantly) get to rails'
+      # String#blank? in SQL
       connection.execute 'TRUNCATE TABLE country_petition_journals'
       connection.execute <<-SQL.strip_heredoc
         INSERT INTO country_petition_journals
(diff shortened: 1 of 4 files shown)


pixeltrix:
Will this make the queries significantly slower?

h-lame:
Probably, but it'll also make them "correct" so I don't know if we have a choice :(.  There's still outstanding discussion about how to manage running these while the site is in use (https://github.com/alphagov/e-petitions/pull/416#issuecomment-164082304 and https://github.com/alphagov/e-petitions/pull/416#issuecomment-164391945).

app/models/country_petition_journal.rb
```diff
@@ -40,6 +42,7 @@ def self.reset!
         timezone('utc', now()), timezone('utc', now())
       FROM signatures
       WHERE state = 'validated'
+      AND country <> ''
```
pixeltrix: Won't this include signatures where country is NULL? Also don't we validate for presence of country so it shouldn't be empty anyway.

h-lame: I guess I was playing safe, our schema allows for null `country` and I'm sure I read that in PSQL `<>` will act as both "NOT NULL" and "disallow blanks".  Could drop the condition though if we felt it was too much.

pixeltrix: Hmm, appears to be true when I just tested it - can you rebase and then I'll merge it, thanks.

h-lame:
rebased.

PR #420, opened 2015-12-17
[WIP] Initial Docker Compose setup

pixeltrix:
Adds basic config for Docker Compose

DO NOT MERGE!

@oskarpearson wdyt? My main concerns for writing the performance tests is that it should be isolated from variables like the email sending and recovery of validation links plus the constituency API.


oskarpearson:
Looks perfect to me :)

pixeltrix:
Closing because I don't think we can really do anything regarding trackable performance statistics

PR #421, opened 2015-12-17
Move email delivery to custom jobs

pixeltrix:
The default email delivery job that comes with Rails 4.2 doesn't do any error handling so create custom email jobs that prevents retries where someone has incorrectly typed their email address and gracefully handles temporary failures so that we don't get Appsignal notifications.


alanth:
LGTM

PR #423, opened 2016-01-02
Add support for queue priorities via named queues

pixeltrix:
Since it's unlikely we'll have budget to add Sidekiq and Redis to the stack until we have to move to the GaaP then we can add this small change so that we don't have to worry about bulk emails not finishing during the evening.

##### Todo

- [x] Merge #421
- [x] Assign jobs to either `:high_priority` or `:low_priority` queues

##### Description

Active Job in Rails 4.2 doesn't support setting a queue priority but does allow setting a queue name. Whilst we could dedicated one worker to processing the validation emails only it would give us more flexibility if the validation emails took priority over the bulk emails for all workers.

To enable this capability this commit adds a before_create callback to the Delayed Job Active Record model that sets the priority of the job based on a predefined mapping of queue names to priorities.

### DIFF
--- a/app/jobs/close_petitions_job.rb
+++ b/app/jobs/close_petitions_job.rb
@@ -1,4 +1,6 @@
 class ClosePetitionsJob < ActiveJob::Base
+  queue_as :high_priority
+
   def perform
     Petition.close_petitions!
   end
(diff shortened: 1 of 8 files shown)


pixeltrix:
@alanth this is ready to go now

h-lame:
Was slightly confused by the potential `queue_as :a_symbol` vs. the `QUEUE_PRIORITIES = { "string_keys" => 0, ...}` thing (even though the tests clearly passed).  Dug through and found that the DJ AJ adapter handles it for us (see: https://github.com/rails/rails/blob/c7644e9eb057bb847009f7d2cbc82dec5cb6c01f/activejob/lib/active_job/queue_adapters/delayed_job_adapter.rb#L17).

Looks good, merging.

PR #424, opened 2016-01-05
Really fix race condition

pixeltrix:
In 97b9839 the find_or_create_by was changed to find_or_initialize_by in an attempt to save a query but unfortunately this introduced a race condition which was supposed to be fixed by 1b07b99. However it wasn't fixed properly because the id would be nil when the query was constructed.

Also refactored the journal code to remove some unnecessary complications.


pixeltrix:
Gah, I know why the tests are failing - the database is stored to millisecond level but the clock resolution on Linux is nanosecond and millisecond on OS X.

alanth:
LGTM

PR #425, opened 2016-01-07
Debate outcome image 105664380

oskarpearson:
Add support for debate outcome photos.

This is quite a lot more complicated than I'd like it to be. There were
- some incompatibilities between Paperclip and our deployment process
- issues with the way that urls are generated when the model is loaded and the asset pipeline
- issues with paperclip and 'hidden' urls - where we put the image behind a proxy
- complexity related to making sure images uploaded are of reasonable sizes

Feedback is welcome!


oskarpearson:
I've done some experimentation after discussion with @pixeltrix - it's much cleaner if we use the fog backend for Paperclip, even although this means we've got two AWS libraries in the system.

- We no longer need the separate Gemfile for deployment and the normal running system
- The paperclip fog backend is more flexible with the handling of alternative hostnames for serving images.

I'm closing this PR, since I don't want to end up with a mangled Git history, and I don't want to force push.

PR #426, opened 2016-01-08
Debate outcome image 105664380 fog

oskarpearson:
Updated image of the voting chamber so that it doesn't appear to always be
completely full. Administrators can now upload a representative image of
the chamber, which will be displayed instead of the placeholder.

It's possible to upload a slightly different image and have the image
automatically resized, but admins will probably get a better looking result if
they upload an image of 1260.0px by 710.0px exactly

The images are stored on S3 and proxied via the webserver at a /attachments/
url, so as to avoid needing separate SSL certificates. We also don't want
browsers to directly fetch files from S3.

Main Changes:

- Added the Paperclip gem to support storage of files on AWS.
- Added a field to DebateOutcome for the associated image.
- Added some validation of the image size and ratios.
- Added tests for the above.
- Included Fog AWS gem, since it's more flexible than the AWS-SDK gem.
  Additionally, Paperclip doesn't currently support AWS-SDK > 2, which
  is required for CodeDeploy, our AWS deployment system.

Other:

- Improved css styling of file upload field.
- Made some changes to the styling of the house of commons image when the
  file sizes are slightly different from expected


oskarpearson:
This replaces https://github.com/alphagov/e-petitions/pull/425 and is made a lot cleaner by using the Fog-AWS backend for Paperclip, rather than AWS-SDK

spec/spec_helper.rb
```diff
@@ -1,4 +1,10 @@
+require 'helpers'
+require 'paperclip/matchers'
+
 RSpec.configure do |config|
+  config.include Helpers
+  config.include Paperclip::Shoulda::Matchers
+
```
pixeltrix: This stuff should be moved to the support folder which will auto-included by rails_helper.rb - something along the lines of this:

``` ruby
# spec/support/paperclip.rb

module PaperclipHelpers
  def commons_image_fixture_path(filename)
    Rails.root.join('spec', 'fixtures', 'images', 'debate_outcome', filename)
  end

  def commons_default_image_url
    ActionController::Base.helpers.image_url('graphics/graphic_house-of-commons.jpg')
  end

  def commons_image_file
    commons_image_fixture_path 'commons_image-2x.jpg'
  end

  def commons_image_file_too_small
    commons_image_fixture_path 'commons_image-too-small.jpg'
  end

  def commons_image_file_wrong_ratio
    commons_image_fixture_path 'commons_image-wrong-ratio.jpg'
  end
end

RSpec.configure do |config|
  config.include PaperclipHelpers
  config.include Paperclip::Shoulda::Matchers
end
```

app/assets/stylesheets/petitions/_forms.scss
```diff
@@ -26,6 +26,10 @@ select.form-control:hover {
 .form-group {
   margin-bottom: $gutter;
   clear: both;
+
+  input[type=file] {
+    line-height: 1ex;
+  }
```
pixeltrix: Is this how they've done file inputs in the various GOV.UK CSS libraries?

app/assets/stylesheets/petitions/views/_shared.scss
```diff
@@ -230,17 +230,17 @@ input.back-page {
...
   &:hover, &:focus {
     opacity: 0.9;
   }
 
+  img {
+    width: 100%;
+    height: 100%;
+    vertical-align: bottom;
+  }
+
```
pixeltrix: What's the vertical-align bottom for? Is it to remove whitespace due to the `` being a inline element? If so then it maybe better to do `display: block` if the `` is the only element inside a `<div>`.

app/controllers/admin/debate_outcomes_controller.rb
```diff
@@ -34,7 +34,8 @@ def fetch_debate_outcome
 
   def debate_outcome_params
     params.require(:debate_outcome).permit(
-      :debated_on, :overview, :transcript_url, :video_url, :debated
+      :debated_on, :overview, :transcript_url, :video_url, :debated,
+      :commons_image
```
pixeltrix: Small nitpick - probably best to leave it all on one line but that just stems from my desktop publishing experience where it was frowned upon to have a line with one word :smile:

app/models/debate_outcome.rb
```diff
@@ -1,10 +1,28 @@
...
+    # default_url needs to be a lambda - this way the generated image url will
+    # include any asset-digest
+    default_url: ->(_) { ActionController::Base.helpers.image_url("graphics/graphic_house-of-commons.jpg") },
+    styles: {
+      "1x": "#{(COMMONS_IMAGE_SIZE[:w]/2).to_i}x#{(COMMONS_IMAGE_SIZE[:h]/2).to_i}",
+      "2x": "#{COMMONS_IMAGE_SIZE[:w]}x#{COMMONS_IMAGE_SIZE[:h]}"
+    }
+
+  validates_attachment_content_type :commons_image, content_type: /\Aimage\/.*\Z/
+  validate :validate_commons_image_dimensions, unless: Proc.new { |a|
```
pixeltrix: Stabby lambda syntax here too

app/models/debate_outcome.rb
```diff
@@ -22,4 +40,28 @@ def date
...
+
+  def validate_commons_image_dimensions
+    # This should be tuned if the images start looking badly scaled
+    max_ratio_delta = 0.1
+
+    dimensions = Paperclip::Geometry.from_file(commons_image.queued_for_write[:original].path)
+
+    # Too big:
+    errors.add(:commons_image, "Width must be at least #{COMMONS_IMAGE_SIZE[:w]}px (is #{dimensions.width}px)") unless dimensions.width >= COMMONS_IMAGE_SIZE[:w]
+    errors.add(:commons_image, "Height must be #{COMMONS_IMAGE_SIZE[:h]}px (is #{dimensions.height}px)") unless dimensions.height >= COMMONS_IMAGE_SIZE[:h]
```
pixeltrix: These lines are too long for trailing conditionals and the errors should be passed as symbols and then looked up from the locale file like other Active Record errors.

app/models/debate_outcome.rb
```diff
@@ -22,4 +40,28 @@ def date
...
+
+    expected_ratio = image_ratio(COMMONS_IMAGE_SIZE[:w], COMMONS_IMAGE_SIZE[:h])
+    actual_ratio = image_ratio(dimensions.width, dimensions.height)
+
+    min_ratio = (expected_ratio - max_ratio_delta).round(2)
+    max_ratio = (expected_ratio + max_ratio_delta).round(2)
+    unless (min_ratio..max_ratio).include? actual_ratio
+      errors.add(:commons_image, "Width and height ratio of uploaded image is #{actual_ratio} - should be between #{min_ratio} and #{max_ratio}")
```
pixeltrix: Again, error message should be looked up by symbol from the locale file

app/views/pages/_home_debated_petitions.html.erb
```diff
@@ -9,6 +9,10 @@
         <% if index == 0 && petition.debate_outcome.video_url? -%>
           <%= link_to petition.debate_outcome.video_url, rel: 'external' do %>
             <span class="debate-video" aria-hidden>
+              <%= image_tag petition.debate_outcome.commons_image.url('2x'),
+                  srcset: "#{petition.debate_outcome.commons_image.url('1x')} 1x, #{petition.debate_outcome.commons_image.url('2x')} 2x",
+                  'aria-hidden': ''
+                  %>
```
pixeltrix: This is probably too much logic here - might want to move it to a view helper especially as it's used in two places

config/initializers/paperclip.rb
```diff
@@ -0,0 +1,22 @@
...
+    fog_credentials: {
+      use_iam_profile: true,
+      provider: 'AWS',
+      region: 'eu-west-1',
+      scheme: 'https'
+    },
+    # Proxied to S3 via the webserver
+    fog_host: '/attachments'
+  )
+end
```
pixeltrix: Any reason this can't be in `config/environments/production.rb` and use `Paperclip.options[]` ?

app/assets/stylesheets/petitions/_forms.scss
oskarpearson: I'm probably not the best person to answer that - https://github.com/alphagov/govuk_elements/tree/master/public/sass/elements/forms doesn't have any file specific input options. Without this, the 'file upload' button is mis-aligned with it's related text. @henryhadlow might be the best person to make a suggestion here?

app/assets/stylesheets/petitions/views/_shared.scss
oskarpearson: Without this and if the image wasn't exactly of the correct proportions, the overlay for the play button would sometimes sit slightly above the bottom of the image.

This worked ok previously because the image was a background - but on converting it to a img the browser would fit the containing div to either the image width or the image height (not both). The exact gap would also change as the page size changed.

I got Henry Turner to double check that I wasn't going crazy here.. `display:block` didn't work for this. There might be a better way. Suggestions welcome.

app/controllers/admin/debate_outcomes_controller.rb
oskarpearson: Fixed. We don't really stick to 80 column width elsewhere, so I've made this the same.

oskarpearson:
@pixeltrix Other than the notes, I think this accounts for everything. Please let me know. Should we wait for @henryhadlow on the CSS side of things? Or is there someone else that could give input?

app/assets/stylesheets/petitions/views/_shared.scss
pixeltrix: The way I'd normally do the play button would be to `position: relative` the wrapper `<div>` and then `position: absolute` the play button with `left: 0; bottom: 0`. But if this works and Henry's okayed it then there's no need to change it.

app/assets/stylesheets/petitions/_forms.scss
henryhadlow: I'm not sure how we usually do file uploads – I don't think we have a recommended pattern. If this works then it's fine.

PR #427, opened 2016-01-12
Switch from a fixed list of disposable domains to dynamic whitelisting/blocking

pixeltrix:
Having a fixed list of disposable domains that can only be changed by deploying makes it very difficult to respond to attacks in a timely manner. So rather than that use a dynamic approach that logs domains and keeps track of any trending domains and then whitelist/block as appropriate. Any domains that are blocked are silently ignored so that a script can't detect easily when it needs to change to a different domain. It logs and aggregates parent domains to provide a defence agains a revolving subdomain attack - blocking the parent domain will block all the subdomains automatically.


h-lame:
Probably makes sense to move the `signature.domain_allowed?` check into the email job to get it out of the request/response cycle in the off chance it's an expensive lookup.  It might be good to introduce a "fraud" state to the signature and if `signature.domain_allowed?` fails in the job we transition the signature - it can also be used as a value to let us find all those signatures later for forensics or de-blocking if we've got it wrong.

Speaking of which, we'll need a script or job to run when an existing domain changes from allowed to blocked or vice-versa.  This would go through signatures that match it and change their states, update counts if they were already validated, actually send the verify email if they were pending.  Perhaps we need fraud-pending and fraud-validated as states to distinguish theses two cases.

pixeltrix:
Closing this because tracking DNS is a thankless task

PR #430, opened 2016-01-21
Use beginning of day for open_at_for_closing

pixeltrix:
Petitions should get the full six months so use the beginning of day when calculating the petitions that need closing. Previously we were effectively closing petitions a day early since they weren't opened at the beginning of the day.


h-lame:
:+1:

PR #434, opened 2016-02-10
Allow editing petition even if petition is already live

henryhadlow:


pixeltrix:
I moved the 'Edit petition' link down to the list of petition actions - wdyt?

henryhadlow:
Ah, I left it next to the petition text to keep it consistent with how it appears in the moderation view. Can we keep it like that?

pixeltrix:
@henryhadlow sure - I'll move it back

PR #435, opened 2016-02-11
Pull locations from country register

pixeltrix:
Use the new country register to populate the location drop down when signing petitions. Some old locations are on the register because they are not officially recognised as countries by the FCO so fill them in manually from the list at https://www.gov.uk/api/world-locations

**IMPORTANT! Until #436 is merged and deployed after this is merged and deployed, country journal data will be partitioned because of the change to use location code rather than company name as the lookup.**


h-lame:
Is there any merit in making the `location_code` fields in `Signature` etc. in the db into foreign keys on the new `Location` table?  Probably after #436 where we should have more complete data.

PR #438, opened 2016-02-12
Use Ruby instead of SQL to migrate data

pixeltrix:
During testing on preview the migration took approximately 40 minutes to run during which time no inserts would be possible. Iterating over all of the rows in Ruby and updating them one by one is much slower but doesn't lock the whole table. Tests on a MacBook Pro showed the migration to take about 4-5 hours but can be left running in the background whilst petitions are still being signed.


oskarpearson:
LGTM.

PR #440, opened 2016-02-14
Retry signature validation if the schema changes

pixeltrix:
When a schema change occurs on the signatures table the prepared statement cache for the lock becomes invalid and the PostgreSQL adapter will raise `PG::InFailedSqlTransaction`. Because the transaction is already invalid any attempt to deallocate the statement fails - see rails/rails#12330.

Whilst any general approach that retries the transaction block after clearing the cache is not a workable solution it's one that can work in this instance. We only retry the once - if clearing the cache doesn't help then the cause of the transaction failure is somewhere else.

### DIFF
--- a/app/models/signature.rb
+++ b/app/models/signature.rb
@@ -104,7 +104,7 @@ class Signature < ActiveRecord::Base
   def validate!
     update_signature_counts = false
 
-    with_lock do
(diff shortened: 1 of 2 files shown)


alanth:
LGTM

PR #442, opened 2016-02-17
Don't load all petitions on the home page

pixeltrix:
The none? method on relations isn't optimised until Rails 5.0 so use the alternative method empty? which is optimised.


oskarpearson:
Looks good to me. Good catch

PR #443, opened 2016-02-17
Fix migrations task for country petition journals

pixeltrix:
A `TRUNCATE` followed by an `INSERT ... FROM` doesn't really work that well on a busy site like the petitions website because new journals will start to appear before the `INSERT` completes. Fix this by doing it in Ruby and locking the journals before calculating the signatures.

To do this in a performant manner we need an index on the petition_id and location_code for the signatures table. We need to do this concurrently otherwise it will block writes to the table, preventing
anyone from signing any petitions.


pixeltrix:
Note that the index will be created manually on production due otherwise it will take a long time to complete a CodeDeploy deployment. Only the changed bit of the migration task will be run on production since the country -> location_code transformation was successful.

alanth:
LGTM

PR #445, opened 2016-02-23
Remove be_valid_asset gem

pixeltrix:
We want to turn off validator.unboxedconsulting.com and it doesn't make sense the tests requiring a network connection to run so remove the `be_valid_asset` gem and replace it with Nokogiri validation. This actually turned up a few problems which have also been fixed.


h-lame:
Turning off validator seems useful and good to me.  However, I'm a little unsure about the markup changes as they suggest we're serving xhtml5, which I don't think we are.  As far as I can tell xhtml5 should start with an XML processing instruction, and be served with an XML mime-type (see: https://html-differences.whatwg.org/#syntax), neither of which is the case for our markup.

pixeltrix:
@h-lame the spec allows you to serve up as `text/html` a well-formed XML document so that you can process it using XML tools. Such a document is said to be using 'polyglot' markup: 
https://dev.w3.org/html5/html-polyglot/html-polyglot.html

h-lame:
Oh, so it does.  But then this version https://www.w3.org/TR/html-polyglot/ of the same spec says it's no longer being worked on :confused:.  Assuming the spec is ok though I think we have to add some `xmlns` attributes to the root `<html>` element (at least that's what my reading of https://www.w3.org/TR/html-polyglot/#element-level-namespaces and https://en.wikipedia.org/wiki/Polyglot_markup#Polyglot_HTML_requirements suggests).

PR #457, opened 2016-04-14
Support new `/records` resource schema from country register

om-sharma:
Response for `/records` resource from country register will be changed very soon. It will start returning new map based response specified in register specification `https://openregister.github.io/specification/#records-resource` rather than current array based response.

  Changes in the PR confirms that the e-petitions is forward compatible with the new-schema from country register. This will keep e-petitions code working for both new and old schema. It also allows country register to publish new response independently.

  Same set of tests are added for the new schema and implemented the solution to support both schemas. The old-schema tests/code can be  deleted after country register starts publishing the new-schema.

### DIFF
--- a/app/jobs/fetch_country_register_job.rb
+++ b/app/jobs/fetch_country_register_job.rb
@@ -26,7 +26,12 @@ class FetchCountryRegisterJob < ActiveJob::Base
   private
 
   def countries
-    fetch_register.body.map{ |r| r['entry'] }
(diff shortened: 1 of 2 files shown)


michaelabenyohai:
This change has now been deployed to the Country Register. It's no longer necessary to support the old response - only support for the response defined at https://openregister.github.io/specification/#records-resource is needed.

PR #461, opened 2016-05-04
Update unsubscribe link to explain up-front why you're receiving emails

henryhadlow:


app/views/petition_mailer/notify_signer_of_debate_scheduled.html.erb
```diff
@@ -1,4 +1,6 @@
-<p>Dear <%= @signature.name %>,</p>
+<p><small>You’re receiving this email because you signed this petition.</small></p>
```
h-lame: @heathd did you mean to remove the "Dear PERSON" part here?  That's what's caused the spec failure.

PR #464, opened 2016-05-10
Remove debate scheduled emails

henryhadlow:
I don’t think they’re relevant


pixeltrix:
Removed the email button so the screen now looks like this:

schedule-debate

pixeltrix:
@henryhadlow rebased against master to fix the merge conflicts

PR #467, opened 2016-05-14
Move tokens from a path param to a query param

### LINKED ISSUE #448: Too many URLs in Google Analytics
because we have the signature ID in the URL

pixeltrix:
When we get more than 50,000+ signatures in a day Google Analytics will send a 'Too many URLs' alert because the tokens are embedded in the path, thus making each one a unique URL.

We can't tell GA to ignore parts of the URL but it can [ignore named query parameters][1], so by moving the token to a query parameter we can ignore it and prevent any future alerts.

Fixes #448.

[1]: https://support.google.com/analytics/answer/1010249?hl=en


h-lame:
This looks good, but it needs a rebase since merging #464 :(

h-lame:
No longer needs a rebase now that #464 has been reverted.  Merging.

PR #469, opened 2016-05-19
Remove debate scheduled emails (revert)

h-lame:
Resurrects #464 (via a revert of #468).

We might want to do this, but at the moment we're not sure the admin users are on board with the idea.  This PR is for discussion and not to be merged until it's clear that we should do it.


h-lame:
Rebased to take into account master having #467 on it which changed some urls.

pixeltrix:
Closing this because there's no demand for it

PR #473, opened 2016-05-27
Add the ability to search by name in the admin area

pixeltrix:
Since we have to preserve the case of the name we can't just lowercase it like we do with the email so to achieve case-insensitiveness we need to create an expression index on the name. Since support for these indexes isn't available until Rails 5.0 we need to use SQL to generate the index and override `index_exists?` because it doesn't appear in the list of indexes returned by Active Record.

https://www.pivotaltracker.com/story/show/120103125

### DIFF
--- a/app/controllers/admin/searches_controller.rb
+++ b/app/controllers/admin/searches_controller.rb
@@ -5,6 +5,8 @@ class Admin::SearchesController < Admin::AdminController
       find_petition_by_id
     elsif query_is_email?
       find_signatures_by_email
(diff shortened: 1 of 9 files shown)


pixeltrix:
Note to search by name you need to add quotes around the name, e.g: "Andrew White"

pixeltrix:
Also note we'll need to create the index manually on production again to prevent CodeDeploy from timing out due to the 15-30 minutes it'll take to create the index.

PR #476, opened 2016-05-29
Fix intermittent spec errors

pixeltrix:
Using the attribute rather than the association when constructing the existing record would sometimes cause duplicate record errors depending on the order of spec execution.


pixeltrix:
Somewhat ironically this failed sporadically but because of a different reason - the phantomjs download sometimes fails because Bitbucket is rate limiting downloads and there's a lot of usage.

PR #482, opened 2016-06-25
Auto-update signature count on petition page.

dracos:
This fetches the JSON of the petition every 10 seconds and updates the count, and progress bar, accordingly, with a pretty counting animation.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -31,6 +31,10 @@ module ApplicationHelper
     params.values_at(:controller, :action) == %w[petitions show]
   end
 
+  def open_petition_page?
+    petition_page? && @petition.open?
+  end
+
   def archived_petition_page?
     params[:controller] == 'archived/petitions' && params[:action] == 'show'
   end
--- a/app/assets/javascripts/auto-updater.js
+++ b/app/assets/javascripts/auto-updater.js
@@ -1,0 +1,54 @@
+//= require jquery.countTo
+//
+// Check for signature count update every few seconds
+// Has hardcoded threshold levels.
+(function ($) {
+  'use strict';
+
+  var JSON_URL = $('.meta-json a').attr('href'),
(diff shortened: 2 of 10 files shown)


app/helpers/application_helper.rb
```diff
@@ -23,6 +23,11 @@ def home_page?
     params.values_at(:controller, :action) == %w[pages index]
   end
 
+  def open_petition_page?
+    # XXX What should go here? Same as petition_page? plus petition.open? Is that allowed here?
+    true
```
pixeltrix: The method `petition_page?` is just a check for the controller and action name but we know that if it's true then the `@petition` instance variable is set so just use the `Petition#open?` method to achieve what you need, e.g:

```
petition_page? && @petition.open?
```

Hopefully that's clearer?

pixeltrix:
The reason that `archived_petition_page` isn't in the fragment.yml file is that the head section isn't dependent on that - it's got `:url` in the list so would change anyway.

app/helpers/application_helper.rb
dracos: Ah, great, thanks, I still don't get my head around where variables are magically available in ruby :)

dracos:
Ah, of course, makes sense, thanks. Think only thing I don't understand now is the `Keys` class in CacheHelper, but I don't really need to to tidy this up ;) Fixes pushed.

pixeltrix:
> Think only thing I don't understand now is the Keys class in CacheHelper

That's just a simple decorator around the template context so that we can return the values in the way that we need. Mostly it's simple delegations but some are slightly more complicated. It then creates the cache key for each item in the list of dependencies.

pixeltrix:
@dracos one more thing - I think creating a signature count API would be useful - there's no need to pull all of the JSON data just to get the signature count.

pixeltrix:
Closing in favour of the fixed #512 - thanks @dracos

PR #484, opened 2016-06-25
Use memcached for petition timestamping

h-lame:
Follow the same pattern that we used for signature_count.  Potential downside here is that we now read 3 things from memcached for a petition instead of one.  I looked at using `fetch_multi`, to get all in one go, but because we need `raw: true` for the `signature_count` it would mean all our values need that and we'd have to remember to translate the values.  Not a huge problem, but it made things a lot more complex than I wanted.

Perhaps we can try this as-is and evaluate it?


h-lame:
Closed as we have merged #491 to master which is a more robust solution for load than those introduced in the production-hotfixes-june-2016 branch - making this pr redundant.

We'll revisit the memcached solution if #491 doesn't prove to be a success.

PR #489, opened 2016-06-26
Calculate GB count for signatures_by_country in json

h-lame:
In 8ae55cc26209e7ac1c7308c607d8cc7553f90af4 we stopped recording the country petition journal entries for 'GB' when people signed.  This meant the data in the JSON api response for a petition looked suspicious as the counts in signatures_by_country did not add up to the total signature_count.  To fix this we calculate the count for GB by subtracting the other country counts from the total count.

We take care to reject any existing GB entry to avoid duplicate entries in the feed.  Note that showing the API will now break if there is no GB Location instance as we'll end up trying to delegate the name of the GB to that instance.

Inserting the GB entry with the correct count is done in a helper method for the api.  Maybe it should be done in Petition#signatures_by_country directly?

### DIFF
--- a/app/helpers/petition_helper.rb
+++ b/app/helpers/petition_helper.rb
@@ -37,6 +37,19 @@ module PetitionHelper
     petition_list_header.present?
   end
 
+  def api_signatures_by_country(petition)
+    signatures_by_country = petition.signatures_by_country
+    # remove 'GB' if it's there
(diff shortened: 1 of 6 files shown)


h-lame:
Closed as we have merged #491 to master which is a more robust solution for load than those introduced in the production-hotfixes-june-2016 branch - making this pr redundant.

PR #490, opened 2016-06-26
Avoid setting state twice in increment_signature_count!

h-lame:
When a petition is collecting sponsors but the creator has not yet validated their own email and we get to the sponsorship threshold we were trying to set the state to both 'validated' and 'sponsored'.  Postgres will complain if you try to set the same column multiple times in a single update and so the update would fail.  Instead of trying to set it twice we make sure to only set it once.

Note that it's not a huge problem if a creator doesn't directly validate their signature.  We know that their email is real if sponsors start coming in as the link for sponsors only appear in the email we send the creator.  If people are hitting that url then we know the email has been delivered so it's ok to go from 'pending' straight to 'sponsored'.

### DIFF
--- a/app/models/petition.rb
+++ b/app/models/petition.rb
@@ -357,12 +357,14 @@ class Petition < ActiveRecord::Base
     updates = []
 
     if pending?
-      updates << "state = 'validated'"
+      updates = ["state = '#{VALIDATED_STATE}'"]
     end
 
     if at_threshold_for_moderation? && collecting_sponsors?
-      updates << "moderation_threshold_reached_at = :now"
-      updates << "state = 'sponsored'"
+      updates = [
+        "moderation_threshold_reached_at = :now",
+        "state = '#{Petition::SPONSORED_STATE}'"
+      ]
     end
 
     if at_threshold_for_response?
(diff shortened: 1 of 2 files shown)


pixeltrix:
Wouldn't be simpler to implement the method like this:

``` ruby
  def increment_signature_count!(time = Time.current)
    updates = []

    if pending?
      updates = ["state = '#{VALIDATED_STATE}'"]
    end

    if at_threshold_for_moderation? && collecting_sponsors?
      updates = [
        "moderation_threshold_reached_at = :now",
        "state = '#{SPONSORED_STATE}'"
      ]
    end

    if at_threshold_for_response?
      updates << "response_threshold_reached_at = :now"
    end

    if at_threshold_for_debate?
      updates << "debate_threshold_reached_at = :now"
      updates << "debate_state = 'awaiting'"
    end

    if updates.size > 0 || update_timestamps?
      updates << "last_signed_at = :now"
      updates << "updated_at = :now"
    end

    if updates.size > 0
      update_all([updates.join(", "), now: time])
    end

    Rails.cache.increment(cached_signature_count_key)
    self.reload
  end
```

h-lame:
I was leaving it open for the other conditionals (or future ones) to also make changes to state, but I guess that's pretty unlikely.  I'll redo.

h-lame:
Closed as included in #491.

PR #491, opened 2016-06-27
Move petition updating to background

h-lame:
Move the calls to `Petition#increment_signature_count!` and writing to the petition journals into a background process.  This allows us to control write contention if we need to.  We also introduce a new priority queue (by demoting existing "high_priority" queue down a notch).

We also include the commit fixing the disposable domains that we did on the production-hotfixes-jun-2016 branch but not on master.


h-lame:
Added a version of the fix in https://github.com/alphagov/e-petitions/pull/490 to here too.

pixeltrix:
LGTM :shipit:

PR #493, opened 2016-06-28
Disallow plus addressing in emails

h-lame:
We'd wanted to leave this in as there are legitimate usecases, but plus addresses make it trivial to sign multiple times from the same email account.

This commit refactors the inline validations from the staged validations into a activemodel validator.


pixeltrix:
LGTM :shipit:

PR #495, opened 2016-06-28
Use typographic quotes and apostrophe in error msg

quis:
> “Smart quotes,” the correct quotation marks and apostrophes, are curly or sloped. "Dumb quotes," or straight quotes, are a vestigial constraint from typewriters when using one key for two different marks helped save space on a keyboard

– http://smartquotesforsmartpeople.com

;)


h-lame:
Thanks @quis!

PR #498, opened 2016-06-29
Catch errors from the Dalli client

pixeltrix:
Sometimes under heavy load the Dalli client is raising exceptions so catch those to return either nil or false.


h-lame:
Good stuff.  Will merge when travis is happy.

PR #501, opened 2016-06-30
Increase signal to noise ratio on exceptions

h-lame:
We get some exceptions that we know we can ignore (MissingTemplate for formats we don't support), or that we only want to happen once not repeatedly (Deserialization errors in the new petition signed data update job caused by deleted signatures).  These commits aim to sort this out.


pixeltrix:
LGTM :shipit:

PR #502, opened 2016-06-30
Retry email jobs when we receive an EOFError

pixeltrix:
Every now and again the SES SMTP gives an EOFError for a variety of reasons. When this happens we should just retry the job.


h-lame:
Good idea.

PR #503, opened 2016-06-30
Fix RSpec warning about false positives

pixeltrix:
Using `not_to raise_error` with a specific class can hide problems with the code causing errors of another kind so follow the recommended practice of not using a specific class.


h-lame:
Oops. I meant to change that in my pr but forgot.

PR #505, opened 2016-06-30
Make sure jobs run at midnight in the application time zone

pixeltrix:
If a petition was published in the winter and then due to close in the summer then using a cron tab where the server time is in UTC would cause a discrepancy in when petitions where closed because it changed the calculated value returned by `Site.opened_at_for_closing`. This resulted in petitions getting an extra days signing time.

To workaround this we run the cron job in the morning and create a scheduled job that runs after midnight and explicit passes the date/time into the job so that we're less reliant on current Date/Time values.


h-lame:
Ugh, timezones.  Looks like the right fix though.

PR #506, opened 2016-07-01
Fix ActiveRecord::RecordNotUnique errors

pixeltrix:
When someone does a double submit we can get uniqueness exceptions from the database. In these circumstances just redirect to the thank you page as the first request will have sent the email.


h-lame:
👍

PR #508, opened 2016-07-07
Add functionality to admin invalidations

pixeltrix:
Currently invalidating signatures is an ad-hoc task but with increasing amounts of fraudulent signatures we need to make it an admin task.

This commit adds an invalidation model that tracks what are the reasons for the invalidation and what are its criteria. It processes them in a background job so that a sysadmin can track the progress of a large invalidation and cancel it if necessary. You can also preview the number of signatures that will be invalidated.

### DIFF
--- a/spec/models/constituency_petition_journal_spec.rb
+++ b/spec/models/constituency_petition_journal_spec.rb
@@ -91,7 +91,7 @@ RSpec.describe ConstituencyPetitionJournal, type: :model do
       end
     end
 
-    context "when the supplied signature is niL" do
+    context "when the supplied signature is nil" do
       let(:signature) { nil }
 
       it "does nothing" do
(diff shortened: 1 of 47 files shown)


spec/models/constituency_petition_journal_spec.rb
```diff
@@ -142,6 +142,95 @@ def journal
...
+      end
+
+      it "updates the updated_at timestamp" do
+        expect {
+          described_class.invalidate_signature_for(signature, now)
+        }.to change { journal.reload.updated_at }.to(now)
+      end
+    end
+
+    context "when the supplied signature is niL" do
```
oskarpearson: nil != niL :)

spec/models/country_petition_jounal_spec.rb
```diff
@@ -143,6 +143,96 @@ def journal
...
+      end
+
+      it "updates the updated_at timestamp" do
+        expect {
+          described_class.invalidate_signature_for(signature, now)
+        }.to change { journal.reload.updated_at }.to(now)
+      end
+    end
+
+    context "when the supplied signature is niL" do
```
oskarpearson: L != l :)

oskarpearson:
Looks great to me.

PR #510, opened 2016-07-14
Add IP rate limiting

pixeltrix:
Trying to block fraudulent signatures by catching disposable domains is like playing a game of whack-a-mole - registering new domains or adding subdomains is trivial and use of wildcard DNS entries allow random generation of subdomains. It's also possible that either the MX record (or the A record when there's no MX record) can point back to a SMTP proxy running on the same computer as the script performing the fraudulent signature task. The only was to try and dynamically block these is by using expensive DNS requests which will likely cause backlogs in the jobs queue when under high load.

However IP addresses are much harder to come by and analysis of the attacks on the 2nd Referendum petition shows that they all came from single IP addresses. Therefore it seems a more efficient solution is to rate limit IP address by the use of two time windows - a short or burst period and a long or sustained period. The default values here are very restrictive to ensure that attackers can't work out the rates from studying this pull request.

Further analysis shows that 90% of our email goes to approximately thirty domains, e.g. gmail.com, outlook.com and broadband providers like sky.com. Since these providers have defences themselves against bulk account misuse we can whitelist them. Also large number of signings came from Wi-Fi providers like The Cloud but their IP addresses are well-known and can be whitelisted also. These whitelists allow us to reduce the IP rate limits to much lower levels to the extent where attackers would need to run scripts for sustained periods of weeks to have any kind of impact.

Depends on PR #508 being merged first.


app/models/rate_limit.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  def ip_whitelist=(value)
+    @ips = nil
+    super(normalize_lines(value))
+  end
+
+  private
+
+  def domains
+    @domains ||= domain_whitelist.each_line.map{ |l| %r[\A#{convert_glob(l.strip)}\z] }
```
oskarpearson: Not a major issue, but it occurred to me that if the user enters blank lines it'll mean that they end up as '\A\z', which is a bit of an odd comparison.

Also, do you think it's worth allowing them to enter '#' as a comment? So they can structure/comment things?

```
# Some situation occurred date XYZ
example.org


# Some other situation occurred here
example.com
```

oskarpearson:
One minor comment on https://github.com/alphagov/e-petitions/pull/510/commits/969ddddadeab75e6a75d68d7c3861c258af8469f - but otherwise looks good to me.

app/models/rate_limit.rb
pixeltrix: Yes, probably a good idea

pixeltrix:
Added blank line and comment filtering

PR #511, opened 2016-07-15
Add task locking

pixeltrix:
Since we run multiple app servers we need to ensure that cron jobs are only executed the once across the autoscaling group. To do this we can add a tasks table that allows us to create a shared lock that works across the whole group.

Depends on #508 and #510 to be merged first.

### DIFF
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -1,0 +1,50 @@
+class Task < ActiveRecord::Base
+  validates :name, presence: true, length: { maximum: 60 }
+
+  class << self
+    def run(name, period = 12.hours, &block)
+      task_for(name).send(:run, period, &block)
+    end
+
+    private
+
+    def task_for(name)
+      begin
+        find_or_create_by!(name: name)
+      rescue ActiveRecord::RecordNotUnique => e
+        retry
+      end
+    end
+  end
+
+  private
+
+  def run(period = 12.hours, &block)
+    retry_lock do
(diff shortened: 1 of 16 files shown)


app/models/task.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def pending?(period)
+    created_at == updated_at || updated_at < period.ago
+  end
+
+  def retry_lock
+    retried = false
+
+    begin
+      with_lock { yield }
```
oskarpearson: According to http://api.rubyonrails.org/classes/ActiveRecord/Locking/Pessimistic.html it seems that 'with_lock' creates a transaction, locks the object, and then runs the yield in the transaction.

Doing something like `FetchCountryRegisterJob.perform_later` in the transaction is probably ok (since it'll execute out-of-band), but `EmailReminder.threshold_email_reminder` (in other parts of this patch) touches a lot of objects and also does 'deliver_now' https://github.com/alphagov/e-petitions/blob/master/app/lib/email_reminder.rb#L6 and https://github.com/alphagov/e-petitions/blob/master/app/lib/email_reminder.rb#L10

I think doing those long running things in a transaction would be problematic?

pixeltrix: Yes, we should convert it to create a background job rather than doing the work inside the block

pixeltrix:
Okay, moved to a background job

PR #512, opened 2016-07-15
Auto update signature count

pixeltrix:
This is a update to #482, with fixes for the following:

* Store current count in a data attribute because of `parseInt`
* Use lightweight signature count API endpoint

### DIFF
--- a/app/assets/javascripts/auto-updater.js
+++ b/app/assets/javascripts/auto-updater.js
@@ -1,0 +1,54 @@
+//= require jquery.countTo
+//
+// Check for signature count update every few seconds
+// Has hardcoded threshold levels.
+(function ($) {
+  'use strict';
+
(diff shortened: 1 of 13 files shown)


app/assets/javascripts/auto-updater.js
```diff
@@ -0,0 +1,56 @@
...
+    } else {
+      $('.signature-count-goal').text(add_commas(THRESHOLD_RESPONSE));
+    }
+  }
+
+  function fetch_count() {
+    $.get(JSON_URL, function(data) {
+      if (data && data.signature_count) {
+        var $count = $('.signature-count-number .count');
+        var current = parseInt($('.signature-count-number .count').data('count'));
```
dracos: This line could presumably use `$count` :)

pixeltrix: Good catch - fixed 👍

PR #519, opened 2016-08-11
Anonymise signatures after 12 months

pixeltrix:
Rather than just deleting the signatures we set them to generic values based on the id of the record and an example postcode for the constituency so that we can still recalculate counts and journals should we wish to retro-fit features onto old petitions that have closed such as signatures by hour.

### DIFF
--- a/app/models/constituency.rb
+++ b/app/models/constituency.rb
@@ -4,6 +4,10 @@ require_dependency 'constituency/api_query'
 class Constituency < ActiveRecord::Base
   MP_URL = "http://www.parliament.uk/biographies/commons"
 
+  MAPIT_HOST = "https://mapit.mysociety.org"
+  MAPIT_AREA_URL = "/area/%{ons_code}"
+  MAPIT_POSTCODE_URL = "https://mapit.mysociety.org/area/%{area_id}/example_postcode"
+
   has_many :signatures, primary_key: :external_id
   has_many :petitions, through: :signatures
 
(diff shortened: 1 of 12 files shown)


app/models/signature.rb
```diff
@@ -177,6 +199,40 @@ def invalidate!(now = Time.current, invalidation_id = nil)
...
+    else
+      self.postcode = nil
+    end
+
+    if postcode.blank? && united_kingdom?
+      # Validations require a postcode for the UK so use the NHS
+      # 'address not known' pseudo-postcode when we didn't find
+      # an example postcode for the constituency:
+      # https://en.wikipedia.org/wiki/Postcodes_in_the_United_Kingdom#ZZ99
+      self.postcode = "ZZ993WZ"
```
h-lame: Oooh, nice.

app/models/constituency.rb
```diff
@@ -46,4 +50,69 @@ def sitting_mp?
...
+    else
+      nil
+    end
+  end
+
+  def fetch_example_postcode(area_id)
+    response = get(MAPIT_POSTCODE_URL % { area_id: area_id })
+
+    if response.success?
+      JSON.load(response.body).upcase.gsub(/\s+/, "")
```
h-lame: Minor, but we could push this through the `PostcodeSanitizer` so it's more obvious what is happening?

app/models/signature.rb
embasbm: Do you need the `self` ?

pixeltrix: @embasbm yes, because it's an attribute on the model - without the `self` it would just create a local variable called `postcode` and not be saved.

embasbm: @pixeltrix great! I just learnt one thing. Thanks

pixeltrix:
Needs to be rethought with the dissolution of parliament

PR #529, opened 2016-08-20
Add the option to geo-block IP addresses

pixeltrix:
If there is a persistent attack using botnets from abroad we want to have the option to block ip addresses from outside the UK.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -36,6 +36,7 @@ gem 'lograge'
 gem 'logstash-logger'
 gem 'jbuilder'
 gem 'paperclip'
+gem 'maxminddb'
 
 # Two AWS libraries:
 #   - aws-sdk v2 for CodeDeploy, which neither Fog nor aws-sdk v1 support
--- a/.env.development
+++ b/.env.development
@@ -7,3 +7,4 @@ MODERATE_HOST=localhost
 SITE_TITLE="Petition parliament (Development)"
 MEMCACHE_SERVERS=localhost:11211
 APPSIGNAL_APP_NAME=epetitions-dev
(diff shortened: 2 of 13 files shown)


Gemfile
```diff
@@ -36,6 +36,7 @@ gem 'lograge'
 gem 'logstash-logger'
 gem 'jbuilder'
 gem 'paperclip'
+gem 'maxminddb'
```
kalleth: IIRC maxminddb needs frequent updating to remain current -- are we going to reflect that with a script on the ops side?

app/assets/stylesheets/petitions/_forms.scss
```diff
@@ -87,3 +87,4 @@ select.form-control:hover {
 .form-hint {
   margin-top: 10px;
 }
+
```
kalleth: churn :nitpick:

Gemfile
pixeltrix: Yes 😉

app/assets/stylesheets/petitions/_forms.scss
pixeltrix: fixed 😄

kalleth:
The code here looks good. :+1: 

[Maxminddb has a 99.8% accuracy rating on countries](https://support.maxmind.com/geoip-faq/geoip2-and-geoip-legacy-databases/how-accurate-are-your-geoip2-and-geoip-legacy-databases/), which might mean we prevent (when this feature has been enabled) 0.2% of the UK from signing permissions.

OTOH, 0.2% of valid submissions not getting through vs preventing an extra 100%+ of false submissions is probably a good tradeoff.

pixeltrix:
The geoipupdate changes have been applied to the preview environment so we're good to merge and test this.

PR #533, opened 2016-08-22
Add trending domains and IP addresses

pixeltrix:
To assist with fraud detection expose the trending domains and IP addresses on the admin home page so that the moderation team can see whether any possible fraud is being attempted. Note that whitelisted entries are removed from the lists because they would likely predominate anyway.

Also add the ability to search by IP address for signatures so that if a trending IP appears the moderation team can check whether all of the signatures are from the same person or different people.


oskarpearson:
Looks good to me

PR #537, opened 2016-09-02
Hide 'At X signatures' blocks if petition is closed

tlfrd:
Hide both 'At 10,000 signatures...' and 'At 100,000 signatures' if a petition is closed

PT: https://www.pivotaltracker.com/n/projects/307301/stories/106483956

### DIFF
--- a/features/suzie_views_a_petition.feature
+++ b/features/suzie_views_a_petition.feature
@@ -94,3 +94,34 @@ Feature: Suzie views a petition
     When I view the petition
     Then I should see "Parliament debated this petition on 26 October 2015"
     And I should see "Waiting for 1 day for Parliament to publish the debate outcome"
+
+  Scenario: Suzie does not see information about future signature targets when viewing a closed petition
+    Given a petition "Spend more money on Defence" has been closed
+    When I view the petition
+    Then I should not see "At 10,000 signatures..."
+    Then I should not see "At 100,000 signatures..."
+
+  Scenario: Suzie sees information about future signature targets when viewing an open petition which has not passed the threshold for response or debate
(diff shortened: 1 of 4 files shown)


features/suzie_views_a_petition.feature
```diff
@@ -94,3 +94,15 @@ Feature: Suzie views a petition
     When I view the petition
     Then I should see "Parliament debated this petition on 26 October 2015"
     And I should see "Waiting for 1 day for Parliament to publish the debate outcome"
+
+  Scenario: Suzie does not see information about future signature targets when viewing a closed petition
+    Given a petition "Spend more money on Defence" has been closed
+    When I view the petition
+    Then I should not see "At 10,000 signatures..."
+    Then I should not see "At 100,000 signatures..."
+
```
pixeltrix: I'd be tempted to add some more scenarios here for when a closed petition has more than 10,000 signatures and more than 100,000 that you don't see those blocks but you do see the response and debate information. Those seem like possible failure scenarios if someone messes up the view logic in future.

features/suzie_views_a_petition.feature
```diff
@@ -94,3 +94,15 @@ Feature: Suzie views a petition
...
+  Scenario: Suzie sees information about future signature targets when viewing an open petition
+    Given an open petition "Spend more money on Defence"
+    When I view the petition
+    Then I should see "At 10,000 signatures..."
+    Then I should see "At 100,000 signatures..."
```
pixeltrix: Again you probably want to add some scenarios for open petitions over 10,000 hiding the first block and then over 100,000 hiding the second block and showing the response and debate information.

PR #545, opened 2016-09-28
Increase allowed size of government responses

pixeltrix:
The limit of 5000 doesn't take into account the fact that posting the form sends CR/LFs which counts as two characters but in Word it's seen as one character by the counter in the document chrome.

Fix this by increasing the validation limit to 6000 characters but still instruct staff to keep responses to 5000 characters.


pixeltrix:
I think the people who are writing the responses are not the people who are entering it so that shouldn't be a problem - they'll come in a word document from a government department.

alanth:
I had to be careful about lf's etc when using the lorem ipsum generator. The count from other apps such as Word probably does a similar thing

PR #546, opened 2016-10-07
Tidy up administration area

pixeltrix:
Improve the code quality, I18n, appearance, testing and security of the administration area - see commit messages for details.


app/controllers/concerns/authentication.rb
```diff
@@ -2,6 +2,7 @@ module Authentication
   extend ActiveSupport::Concern
 
   included do
+    before_action :require_admin_and_check_for_password_change
```
kalleth: nitpick: Is this Authentication or Authorisation?

kalleth: nitpick: Is this authentication or authorisation?

app/views/application/_messages.html.erb
```diff
@@ -1,6 +1,3 @@
 <% if flash[:notice] %>
   <p class="flash-notice"><%= flash[:notice] %></p>
 <% end %>
-<% if flash[:error] %>
```
kalleth: Does this harm anything? As a developer working on the application if flash messages had the ability to be displayed, I'd expect the standard categories to display without too much extra work (error, notice, warning). Removing this feels like it'd just surprise Future Me.

app/controllers/concerns/authentication.rb
```diff
@@ -27,24 +27,21 @@ def redirect_to_target_or_default
 
   def require_admin
     unless current_user
-      flash[:error] = "You must be logged in as an administrator to view this page."
-      redirect_to admin_login_url
+      redirect_to admin_login_url, alert: "You must be logged in as an administrator to view this page."
```
kalleth: nit: feels like this commit (which did use `flash[:error]`) and the last commit are out of sequence, because at that point the app *did* use flash[:error].

app/controllers/admin/user_sessions_controller.rb
```diff
@@ -1,5 +1,6 @@
 class Admin::UserSessionsController < Admin::AdminController
-  skip_before_filter :require_admin_and_check_for_password_change
+  skip_before_filter :require_admin, only: [:new, :create]
+  skip_before_filter :check_for_password_change
```
kalleth: This doesn't make much sense to me. It seems like instead of checking if the user needs to change their password on each login, which would seem "Usual (tm)" to me, we're checking on each pageview *except* when a user creates a session.

Perhaps I'm missing a requirement?

app/controllers/admin/admin_users_controller.rb
```diff
@@ -1,52 +1,58 @@
...
     @user = AdminUser.new
   end
 
   def create
-    @user = AdminUser.create(admin_user_params)
+    @user = AdminUser.new(admin_user_params)
+
     if @user.save
-      redirect_to admin_admin_users_url, notice: "User was successfully created"
+      redirect_to admin_admin_users_url, notice: :user_created
```
kalleth: So I'm curious -- given there's only a single locale present in `config/locales` and this is the admin interface for the petitions system for the UK government, is this premature?

OTOH, now it's done, I don't feel strongly enough about it to say whether it's good or bad.

app/controllers/concerns/flash_i18n.rb
```diff
@@ -0,0 +1,30 @@
+module FlashI18n
+  protected
+
+  def redirect_to(url, options = {})
+    self.class._flash_types.each do |flash_type|
```
kalleth: And here's where I get a little bit confused about the necessity for the translations. If there's a user need driving it, then ignore everything I'm about to say;

This "feels railsy" enough to confuse new developers into thinking it's a standard rails feature, but it isn't, so it feels dangerous to be doing so -- adding a layer of indirection that isn't immediately evident to the developer (because they'll think "Neat, i didn't know rails could do that") and proceed past, without realising there's code here that could be causing a bug they're seeing.

app/controllers/concerns/flash_render.rb
```diff
@@ -0,0 +1,23 @@
+module FlashRender
+  extend ActiveSupport::Concern
+
+  include FlashI18n
+
+  def render(options = {}, locals = {}, &block)
```
kalleth: As above

app/views/admin/admin_users/index.html.erb
```diff
@@ -14,7 +14,13 @@
       <td><%= mail_to user.email %></td>
       <td><%= user.role %></td>
       <td><%= user.account_disabled ? 'Yes' : '' %></td>
-      <td class="action last"><%= cms_delete_link user, :url => admin_admin_user_path(user) %></td>
+      <td class="action last">
+        <% if user != current_user %>
+          <%= cms_delete_link user, :url => admin_admin_user_path(user) %>
+        <% else %>
+          &nbsp;
```
kalleth: O_o

IANAFrontenddeveloper but this seems a bit hacky?

app/models/admin_user.rb
```diff
@@ -23,11 +27,54 @@ class AdminUser < ActiveRecord::Base
...
+    elsif !valid_password?(current_password)
+      errors.add(:current_password, :invalid)
+    elsif current_password == password
+      errors.add(:password, :taken)
+    end
+
+    errors.empty? && save(validate: false)
+  end
+
+  def destroy(current_user: nil)
```
kalleth: I've always been wary about overriding methods provided by rails (and then calling super) on ActiveRecord objects as that's historically been the source of some unexpected behaviour.

Could we use a `before_destroy` hook here, or alternatively, move the destroy logic into an action/command/service object (if that's a pattern already in use in this project) to keep the model light?

app/models/admin_user.rb
```diff
@@ -23,11 +27,54 @@ class AdminUser < ActiveRecord::Base
...
   # = Finders =
   scope :by_name, -> { order(:last_name, :first_name) }
   scope :by_role, ->(role) { where(role: role).order(:id) }
 
   # = Methods =
+  def current_password
+    defined?(@current_password) ? @current_password : nil
+  end
+
+  def current_password=(value)
```
kalleth: Could we replace this method (and the one above) with `attr_accessor :current_password` ?

app/models/admin_user.rb
```diff
@@ -23,11 +27,54 @@ class AdminUser < ActiveRecord::Base
...
+    @current_password = value
+  end
+
+  def update_with_password(attrs)
```
kalleth: Similar comment to the below. I'm a fan of thin models used as a data access layer with business logic held elsewhere in a structured manner so I'd be looking to extract this outside the model.

However, if it's not a pattern present already in this project, then I guess it could be kept here, or moved to a concern.

app/views/application/_messages.html.erb
pixeltrix: The key `:error` isn't one of the standard categories, if we wanted to be future proof then you'd do something like this:

``` ruby
<% flash.each.key do |key, value| %>
  <%= content_tag(:p, value, class: "flash-#{key}" %>
<% end %>
```

But you know, YAGNI.

app/controllers/concerns/authentication.rb
pixeltrix: The `:error` was in the admin, not in the front end which has a different layout.

app/controllers/concerns/authentication.rb
pixeltrix: Bit of both - it's what it's been historically called but I'm happy to rename.

app/controllers/concerns/authentication.rb
kalleth: :+1:

app/views/application/_messages.html.erb
kalleth: And here's where your rails knowledge gets the best of me :) I've been using :error, but it appears rails expects :alert. Cool.

app/controllers/admin/admin_users_controller.rb
pixeltrix: There was talk of Welsh at some point and we got an email from Montreal saying they wanted to use the codebase which being in Quebec, Canada needs French and English.

app/views/admin/admin_users/index.html.erb
pixeltrix: Just hides the delete button on the user row that represents the current user - not really front end dev. What would you suggest instead?

app/controllers/concerns/flash_render.rb
pixeltrix: Ditto :smile:

app/controllers/admin/admin_users_controller.rb
kalleth: Gotcha :+1:

app/controllers/concerns/authentication.rb
kalleth: Probably easiest to leave as-is then.

app/models/admin_user.rb
pixeltrix: This is a similar pattern to what Devise uses to update a password on profile page - we don't have a 'services' folder so I'd be very against adding a "PasswordUpdateService" class.

app/models/admin_user.rb
pixeltrix: Possibly - I'll check it out

app/views/admin/admin_users/index.html.erb
kalleth: I'm more talking about the addition of the &nbsp;. Kind of a nit. I'd be hoping that there'd be a CSS way for the frontend to render acceptably if the button element wasn't present?

app/models/admin_user.rb
pixeltrix: Again no other service objects exist so don't want to introduce one now and `before_destroy` didn't really work when I was looking at this. I understand your concern about overriding `destroy` but that stems from the days of `alias_method_chain` - `super` is how we override things in the Rails framework itself now, e.g. the ActiveModel::Validations module.

app/views/admin/admin_users/index.html.erb
pixeltrix: This maybe my age showing but I remember Mozilla/Firefox messing up table layout if a table cell had no content so probably paranoia on my part.

app/controllers/concerns/flash_i18n.rb
kalleth: I'm... not wild about it living in the app codebase. But in general I'm not wild about overriding library/framework methods -- I prefer wrapping them so it's immediately obvious to the developer that they're custom methods.

However, you're the one that'll be working on it with the rest of the epets team, so it's ultimately your call :)

app/models/admin_user.rb
kalleth: I agree with not implementing a service object pattern in the app just for this, so happy to let this go through, because you pinkie promised to keep an eye on the `AdminUser` class and make sure it doesn't get too big :)

PR #554, opened 2017-02-10
Add an all local petitions view

pixeltrix:
We get a lot of requests for top petitions in a constituency but the current page only shows open petitions. To reduce the need for running custom database queries we can add a page to show all petitions and then direct any FoI requests to the page for the constituency.

### DIFF
--- a/app/models/petition.rb
+++ b/app/models/petition.rb
@@ -262,16 +262,16 @@ class Petition < ActiveRecord::Base
       where(id: Signature.petition_ids_with_invalid_signature_counts).to_a
     end
 
-    def popular_in_constituency(constituency_id, how_many = 50)
+    def popular_in_constituency(constituency_id, count = 50)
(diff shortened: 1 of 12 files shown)


app/models/petition.rb
```diff
@@ -293,6 +293,17 @@ def mark_petitions_as_debated!(date = Date.current)
...
+    def popular_in(constituency_id, count)
+      klass = ConstituencyPetitionJournal
+      constituency_signature_count = klass.arel_table[:signature_count].as('constituency_signature_count')
+      constituency_signatures_for = klass.with_signatures_for(constituency_id).ordered
+
+      select(arel_table[Arel.star], constituency_signature_count).
+      joins(:constituency_petition_journals).
+      merge(constituency_signatures_for).
+      limit(count)
+    end
```
h-lame: This is definitely better than the string interpolation version we had before. :clap:

PR #555, opened 2017-02-11
Improve local petition result pages

pixeltrix:
A view for all popular local petitions was adding in #554 to help address FoI requests. It's likely that those requests will want it in data form too so provide an option to get the data in JSON or CSV format.

Also wrap some fragment caching around all the format response to ensure that the database doesn't become overloaded when responding to requests for popular local petitions.


yahoopete:
:+1:

PR #560, opened 2017-03-09
Add GMail action buttons

pixeltrix:
Allow one-click actions to sign or unsubscribe.

To complete this we need to go through the whitelisting process as described here:

https://developers.google.com/gmail/markup/registering-with-google

As far as I can see we're in good shape with the guidelines but it's hard to tell on the more subjective ones like "A very very low rate of spam complaints from users".

We need to send the emails to the whitelist address from the production servers and the easiest way to do this is deploy and signup the address to a petition. That will work for the 'Sign Petition' action but we'll need to manually validate and send another email with the 'Unsubscribe' action.


pixeltrix:
Closing because Google needs to approve these for them to work and it seems like too much effort for any perceived benefit.

Issue #458, opened 2016-04-17
It would be cool to publish the 'Trending petitions' history as open data


henryhadlow:
Proposal:
- Add journaling per petition to record the signature count hourly
- To keep it performant you could have daily rows with 24 columns, one for each hour
- The signing rate could be easily calculated, just the difference between one hour's count and the previous count
- We could then add this data to the existing json for each petition (eg https://petition.parliament.uk/petitions/124333.json)

This would let people see the signing history of petitions and make mashups where you could see trending petitions against events in the news

PR #561, opened 2017-03-18
Add trending petition journals

cih:
This change adds trending petition journals for recording signature
counts at hourly intervals and returns them from the petition show
API endpoint as suggested in #458.

### DIFF
--- a/app/helpers/signature_helper.rb
+++ b/app/helpers/signature_helper.rb
@@ -11,6 +11,10 @@ module SignatureHelper
     t(:"#{key}.html", options)
   end
 
+  def signatures_by_hour(petition)
+    TrendingSignatureCollection.new(petition).hourly_intervals
+  end
+
   private
 
   def render_signature_hidden_details(stage_manager, form, options)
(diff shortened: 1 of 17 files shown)


pixeltrix:
@cih you can rebase against master and force push to rebuild this and it should work

dom96:
Why is this stalled?

pixeltrix:
@dom96 because the journal tables have always been a pain when we get a massive spike in demand like we saw with the Revoke Article 50 petition and the thought of adding more journals fills me with dread at the thought of having to resync them all. Not closed to the idea of providing this data just looking to see if there's a way we can make PostgreSQL do the heavy lifting.

CodePint:
Heya, just thought I would share the project i'm working on to add full tracking to petitions by geography/locale.

https://github.com/CodePint/PetitionTrackerUK

Bulk of the work is done, but without optimisations you are correct that regular polling adds a large amount of DB load, I am working on a solution to this currently.

Likely solution will be:
-  Base scenario of only creating a new record every hour for all petitions above 500 signatures
- Maintain a table of trending petitions which will be polled more regularly
- Some form of dataset reduction and archival after a month.

PR #562, opened 2017-03-31
Update to handle new Register API response format

samcrang:
The response you receive from a request for the `/records` endpoint on a Register is changing. This will be a breaking change. This PR provides support for both the old and new response style whilst we transition.

We're looking to publish an [ADR](https://github.com/openregister/openregister-java/tree/master/doc/arch) which describes the rationable for the breaking change. 

We should be deploying this change soon. We'll send a second PR to remove the code to handle the old format after we've deployed.


samcrang:
I'm assuming the build is failing because I can't trigger Travis builds with access to the necessary secrets and whatnot?

pixeltrix:
> I'm assuming the build is failing because I can't trigger Travis builds with access to the necessary secrets and whatnot?

Yes, we have our own copy of phantomjs in a private S3 bucket due to rate limiting on the bitbucket site and Travis CI doesn't decrypt the environment vars for PRs since that could be exploited to reveal the secret. I'd welcome suggestions on how to fix that without opening up the download to the whole world. 😄

samcrang:
Makes sense. I'm assuming there's a reason you don't just run `rake` and [let the PhantomJS gem](https://github.com/alphagov/e-petitions/blob/master/features/support/custom_env.rb#L14) install things? This is worked locally for me.

pixeltrix:
> Makes sense. I'm assuming there's a reason you don't just run rake and let the PhantomJS gem install things? This worked locally for me.

Yes - it's all documented in ariya/phantomjs#13953. I'm trying to just rely on the phantomjs in the path in PR #563.

pixeltrix:
@samcrang I've merged #563 - can you try rebasing and force pushing to update the PR, thanks.

samcrang:
Build has gone green! 😄

pixeltrix:
@samcrang thanks! 👍 

When is the new format going live?

samcrang:
Thanks!

Hopefully Monday, 3rd April. We'll issue a clean-up PR after we've done it if that helps you out?

pixeltrix:
> Hopefully Monday, 3rd April.

Better get a deploy out then, hadn't I 😄 

> We'll issue a clean-up PR after we've done it if that helps you out?

That'd be great, thanks 👍

samcrang:
Ha, totally forgot you'd still have to deploy 😄. Sorry about the super-late notice on this.

samcrang:
@pixeltrix did the deploy go okay? Are we good to deploy our side on Monday?

pixeltrix:
Yes, we deployed

PR #565, opened 2017-04-01
Update specs using opened_at_for_closing in the past

cih:
Due to the special cases in the Site#opened_at_for_closing method when
the specs are run on the day after a special case 1.day.ago being passed
to the method results in the same time being returned as for the current
time which causes no petitions to be returned from the
Petition#in_need_of_closing method.


pixeltrix:
Ahh - so the specs failed yesterday. 👍

PR #567, opened 2017-04-19
Standardise pagination for archived petitions

AbraaoMota:
The page for Archived petitions was using an inconsistent styling for
pagination in comparison to the active petitions page.
The contents of the paginated results were extracted into a shared
partial where both pages can pull their data from.


app/helpers/search_helper.rb
```diff
@@ -1,12 +1,13 @@
 module SearchHelper
   def paginate(petitions)
+    petition_class = petitions.klass.first.class
```
pixeltrix: This will trigger an SQL query if the relation hasn't been loaded and the relation returned by `klass` has it's own `klass` method that will return the Active Record model without a query. I'd suggest adding `model` method to `Browseable::Search` that returns `klass.klass` and then you don't need the local variable and you can change the lines below to:

``` ruby
previous_link: polymorphic_path(petitions.model, petitions.previous_params),
next_link: polymorphic_path(petitions.model, petitions.next_params)
```

pixeltrix:
Other than the one comment 👍  - you need to check why the test failed though

PR #568, opened 2017-04-19
Add Parliament model for to handle dissolution

pixeltrix:
This could've been handle with hard-coded text and a constant for the date but things aren't completely settled yet so add a model that will give us some flexibility. It will also probably be needed once a new parliament is sitting to split petitions into their respective parliaments.

### DIFF
--- a/app/models/parliament.rb
+++ b/app/models/parliament.rb
@@ -1,0 +1,43 @@
+class Parliament < ActiveRecord::Base
+  class << self
+    def before_remove_const
+      Thread.current[:__parliament__] = nil
+    end
+
+    def instance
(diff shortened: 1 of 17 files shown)


app/models/parliament.rb
```diff
@@ -0,0 +1,43 @@
...
+    def reload
+      Thread.current[:__parliament__] = nil
+    end
+
+    def last_or_create
+      order(created_at: :desc).first_or_create
+    end
+  end
+
+  def dissolved?(now = Time.current)
```
FriedSock: Cant see why you would pass a different time into this method (and its corresponding class method), id rather remove it than pre-empt a use case.

PR #569, opened 2017-04-20
Make pagination helper return html

AbraaoMota:
The `paginate` helper method was appending its output through a side
effect of `concat`. It now returns its html, and the view was updated
accordingly.

### DIFF
--- a/spec/helpers/search_helper_spec.rb
+++ b/spec/helpers/search_helper_spec.rb
@@ -1,64 +1,98 @@
 require 'rails_helper'
 
 RSpec.describe SearchHelper, type: :helper do
-
   describe "#paginate" do
+    let(:page_stubs) do
+      {
+        total_pages: 3,
+        previous_link: "/petitions?page=1&state=all",
+        next_link: "/petitions?state=all",
+        model: Petition
+      }
+    end
 
-    let (:page_params) {{
-      :total_pages=>3,
-      :previous_link=>"/petitions?page=1&state=all",
-      :next_link=>"/petitions?state=all",
-      "model" => Petition
-    }}
+    context "when viewing the first page of the petitions search results" do
+      let(:first_page_stubs) do
+        {
+          first_page?: true,
+          last_page?: false,
+          previous_params: { state: :all, page: nil },
+          next_params: { state: :all, page: 2 },
+          previous_page: nil,
+          next_page: 2,
+        }
+      end
 
-    let (:first_page_params) {{
-      "first_page?" => true,
-      "last_page?" => false,
-      "previous_params" => {:state=>:all, :page=>nil},
-      "next_params" => {:state=>:all, :page=>2},
-      :previous_page=>nil,
-      :next_page=>2,
-    }}
+      let(:petitions) { double('petitions', page_stubs.merge(first_page_stubs)) }
 
-    let (:intermediary_page_params) {{
-      "first_page?" => false,
-      "last_page?" => false,
-      "previous_params" => {:state=>:all, :page=>1},
-      "next_params" => {:state=>:all, :page=>3},
-      :previous_page=>1,
-      :next_page=>2,
-    }}
-
-    let (:last_page_params) {{
-      "first_page?" => false,
-      "last_page?" => true,
-      "previous_params" => {:state=>:all, :page=>2},
-      "next_params" => {:state=>:all, :page=>nil},
-      :previous_page=>2,
-      :next_page=>nil,
-    }}
-
-    context "when it handles the first page of petitions" do
-      it "only adds a link to the next page" do
-        petitions = double('petitions', page_params.merge(first_page_params))
+      it "adds a link to the next page" do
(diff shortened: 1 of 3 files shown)


spec/helpers/search_helper_spec.rb
```diff
@@ -1,6 +1,68 @@
 require 'rails_helper'
 
 RSpec.describe SearchHelper, type: :helper do
+
```
pixeltrix: Remove this whitespace

spec/helpers/search_helper_spec.rb
```diff
@@ -1,6 +1,68 @@
...
+  describe "#paginate" do
+
```
pixeltrix: Remove this whitespace

spec/helpers/search_helper_spec.rb
```diff
@@ -1,6 +1,68 @@
...
+    let (:page_params) {{
+      :total_pages=>3,
+      :previous_link=>"/petitions?page=1&state=all",
+      :next_link=>"/petitions?state=all",
+      "model" => Petition
+    }}
```
pixeltrix: My personal preference is to use `do ... end` for multi-line `let` blocks and I think that's what we've used elsewhere in the project. When making a PR, especially to a public project, you should try to adapt your code style to that of the project's even when it isn't enforce via rubicop 😄

pixeltrix: Also try to use modern hash syntax, e.g. `{ foo: "bar" }` over `{ :foo => "bar" }`

spec/helpers/search_helper_spec.rb
```diff
@@ -1,6 +1,68 @@
...
+      "next_params" => {:state=>:all, :page=>nil},
+      :previous_page=>2,
+      :next_page=>nil,
+    }}
+
+    context "when it handles the first page of petitions" do
+      it "only adds a link to the next page" do
+        petitions = double('petitions', page_params.merge(first_page_params))
+        expect(paginate(petitions)).to include "Next"
+        expect(paginate(petitions)).not_to include "Previous"
```
pixeltrix: Probably want to add some expectations on the numbers too

spec/helpers/search_helper_spec.rb
```diff
@@ -1,6 +1,68 @@
...
+      :next_page=>2,
+    }}
+
+    let (:last_page_params) {{
+      "first_page?" => false,
+      "last_page?" => true,
+      "previous_params" => {:state=>:all, :page=>2},
+      "next_params" => {:state=>:all, :page=>nil},
+      :previous_page=>2,
+      :next_page=>nil,
```
pixeltrix: Why the mix of symbol and string keys in the hashes?

spec/helpers/search_helper_spec.rb
```diff
@@ -1,6 +1,68 @@
...
+    let (:first_page_params) {{
+      "first_page?" => true,
+      "last_page?" => false,
+      "previous_params" => {:state=>:all, :page=>nil},
+      "next_params" => {:state=>:all, :page=>2},
+      :previous_page=>nil,
+      :next_page=>2,
+    }}
+
+    let (:intermediary_page_params) {{
```
pixeltrix: I think calling this params is misleading - they're stubbed methods on the petitions double. I'd prefer to see them inside the context block for each case so that they're grouped together lexically and not composed which masks what the actual stubs are.

spec/helpers/search_helper_spec.rb
```diff
@@ -1,6 +1,68 @@
...
+    let (:last_page_params) {{
+      "first_page?" => false,
+      "last_page?" => true,
+      "previous_params" => {:state=>:all, :page=>2},
+      "next_params" => {:state=>:all, :page=>nil},
+      :previous_page=>2,
+      :next_page=>nil,
+    }}
+
+    context "when it handles the first page of petitions" do
```
pixeltrix: A better description would be "when viewing the first page of the petitions search results" - same goes for the other descriptions.

spec/helpers/search_helper_spec.rb
```diff
@@ -1,6 +1,68 @@
...
+      it "only adds a link to the next page" do
+        petitions = double('petitions', page_params.merge(first_page_params))
+        expect(paginate(petitions)).to include "Next"
+        expect(paginate(petitions)).not_to include "Previous"
+      end
+    end
+
+    context "when it handles an intermediary page of petitions" do
+      it "adds both pagination links" do
+        petitions = double('petitions', page_params.merge(intermediary_page_params))
```
pixeltrix: Move the petitions to a `let` block inside the `context` and then split the expectations into two parts - "adds a link to the previous page" and "adds a link to the next page". The idiomatic form for RSpec is a single expectation per example though obviously that's flexible where needs must.

pixeltrix: Same applies to the other examples and also to the ones you add for expectations on the numbers

PR #570, opened 2017-04-20
Automatically close petitions after dissolution

pixeltrix:


pixeltrix:
Closing this since we're closing petitions because of dissolution at a different time to normal

PR #571, opened 2017-04-22
Additional changes require for the dissolution of parliament

pixeltrix:
* Adds admin CRUD for updating the parliament instance
* Adds a dissolution heading since it's no longer at midnight so needs to be editable
* Updates the existing early closing email to the latest copy
* Adds a button to the admin CRUD to send the early closing email
* Adds a job for closing petitions early at the scheduled dissolution date/time
* Adds a button to the admin CRUD to schedule the job to close petitions early

### DIFF
--- a/app/models/parliament.rb
+++ b/app/models/parliament.rb
@@ -12,6 +12,10 @@ class Parliament < ActiveRecord::Base
       instance.dissolution_at
     end
 
+    def dissolution_heading
+      instance.dissolution_heading
+    end
+
     def dissolution_message
       instance.dissolution_message
     end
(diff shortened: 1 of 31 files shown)


app/models/parliament.rb
```diff
@@ -33,8 +37,14 @@ def last_or_create
     end
   end
 
+  validates_presence_of :dissolution_heading, :dissolution_message, if: :dissolution_at?
+  validates_length_of :dissolution_heading, maximum: 100
+  validates_length_of :dissolution_message, maximum: 600
+
+  after_save { Site.touch }
```
cih: Is this so that we bust the petition fragment cache? If so are we intentionally busting that when the site gets updated or could we use the parliament `updated_at` timestamp?

pixeltrix: Possibly but it also busts the head cache which contains the application.css link

pixeltrix:
Thanks for the review 👍

PR #572, opened 2017-04-24
Allow specifiying an FAQ link when dissolving parliament

h-lame:
If present this link will be displayed at the bottom of the message as follows:

> Find out more on the **Petitions Committee website**

Where the part in bold will be the clickable link.  If no url is provided that sentence is not included in the warning notification at all.

### DIFF
--- a/app/views/application/_parliament_dissolution_warning.html.erb
+++ b/app/views/application/_parliament_dissolution_warning.html.erb
@@ -3,5 +3,8 @@
     <span class="icon icon-warning-white"></span>
     <h3 class="header"><%= Parliament.dissolution_heading %></h3>
     <p class="content"><%= Parliament.dissolution_message %></p>
+    <% if Parliament.dissolution_faq_url? %>
+      <p class="content">Find out more on the <%= link_to 'Petitions Committee website', Parliament.dissolution_faq_url %></p>
+    <% end %>
   </div>
 <% end %>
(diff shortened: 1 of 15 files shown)


app/views/application/_parliament_dissolution_warning.html.erb
```diff
@@ -0,0 +1,10 @@
+<% if Parliament.dissolution_announced? %>
+  <div class="notification">
+    <span class="icon icon-warning-white"></span>
+    <h3 class="header"><%= Parliament.dissolution_heading %></h3>
+    <p class="content"><%= Parliament.dissolution_message %></p>
+    <% if Parliament.dissolution_faq_url.present? %>
```
pixeltrix: The `.present?` is unnecessary - the query attribute method is enough

pixeltrix:
:shipit:  the present change isn't worth bothering with

app/views/application/_parliament_dissolution_warning.html.erb
h-lame: I always forget about those. I have a mental block on them.

h-lame:
Made the query attribute change, also added `:site_updated_at` to the fragments used to invalidate the cache on the homepage.  Without this changing the Parliament model wouldn't refresh the home page block where the "parliament dissolved" notification now lives.

PR #573, opened 2017-04-28
Remove duplicated signature warning

AbraaoMota:
When attempting to sign a petition, there is a validation step ran in
the form submission to prevent people from signing the petition multiple
times using the same email. This appears in the form of an error
saying "You cannot sign this petition again". This presents a privacy
concern to users and leaks information about which emails (and email
owners) have signed what petitions.

This commit mitigates this by delaying the validation step into the
controller checks when creating a signature. When attempting to sign
a petition a second time (using the same details), the owner of the
email will receive an email alerting them their email was used in a
duplicated signature attempt. To prevent spamming inboxes, an
`email_count` attribute has been added to the signature table to keep a
track of emails sent to the owner of the email. The duplicate signature
email as well as the confirmation email sent out for the first time will
both increase this counter, and neither will be sent out when the
counter hits 5.


db/migrate/20170421153056_add_email_count_to_signatures.rb
```diff
@@ -0,0 +1,5 @@
+class AddEmailCountToSignatures < ActiveRecord::Migration
+  def change
+    add_column :signatures, :email_count, :integer
```
h-lame: If we add a default 0 here we can drop the `def email_count; super || 0; end` in `Signature`

h-lame: There may be performance issues with doing this though, not sure if adding default 0 to this column might take ages given how big this table is.

pixeltrix: @h-lame yes, adding 0 as the default would lock the whole table during migration whilst it backfills 30 million rows. However since we're emptying the whole table shortly that would be a good time to add it.

app/models/signature.rb
```diff
@@ -266,6 +265,29 @@ def rate(window = 5.minutes)
...
+    return false if existing_email_address_count == 0
+    return true if existing_email_address_count > 1
+    existing_signature = matcher.first
+    return true if (existing_signature.name.strip.downcase == name.strip.downcase)
+    if (existing_signature.postcode.gsub(/\s+/,'').downcase !=
+        postcode.gsub(/\s+/,'').downcase)
+      return true
+    end
+    false
+  end
```
h-lame: Total nitpick but I think the last few lines here (after `existing_signature = matcher.first`) could be made clearer - we've mixed the styles guard clause vs. full if statement and I think it would be clearer if it was all one style so either:

```ruby
return true if (existing_signature.name.strip.downcase == name.strip.downcase)
return true if (existing_signature.postcode.gsub(/\s+/,'').downcase != postcode.gsub(/\s+/,'').downcase)
return false
```

or:

```ruby
if (existing_signature.name.strip.downcase == name.strip.downcase ||
   (existing_signature.postcode.gsub(/\s+/,'').downcase != postcode.gsub(/\s+/,'').downcase)
  true
else
  false
end
```

Or some 3rd option that you feel works.

db/migrate/20170421153056_add_email_count_to_signatures.rb
AbraaoMota: So should I add the default 0 for this PR? @pixeltrix

pixeltrix: @AbraaoMota on balance I think yes, since we won't merge it until we've archived the petitions and signatures.

pixeltrix:
Closing because the changes are incorporated as part of #622

PR #586, opened 2017-05-08
Refactor example postcode lookup

pixeltrix:
MapIt's example postcode returns a random postcode and sometimes this is a the edge of the constituency boundary. This leads to disagreements between Parliament's API and MapIt as to which constituency a postcode belongs to.

To try and fix this use an alternative method to get the nearest postcode to the centroid of the constituency boundary. This is also subject to possible errors where the centroid lies within a 'hole' in the constituency so we add a check for that just in case.


pixeltrix:
Having looked at some of the boundaries this will almost certainly fail

PR #594, opened 2017-06-10
Add register to vote link back

pixeltrix:
Switch to get involved link automatically based on the date and time that registration closes for an election.


spec/models/parliament_spec.rb
```diff
@@ -270,4 +276,36 @@
...
+      end
+    end
+
+    context "when registration_closed_at is in the future" do
+      subject :parliament do
+        FactoryGirl.create(:parliament, :dissolving, registration_closed_at: 2.weeks.from_now)
+      end
+
+      it "returns false" do
+        expect(parliament.dissolved?).to eq(false)
```
NeilvB: I think this and the spec below should be calling `#registration_closed?`

PR #595, opened 2017-06-12
Switch to using a namespace module for archiving

pixeltrix:
Since we will be adding extra models for archived petitions like debate outcomes and government responses switch to using a module for namespacing.


NeilvB:
Looks good to me 👍

PR #596, opened 2017-06-14
Add migration task for legacy data

pixeltrix:
Before we can archive existing petition data we need to migrate the existing archived petitions to the new schema. Unfortunately we don't have enough information to accurately determine a number of timestamps so we make educated guesses so that we can setup facets for the various lists of petitions.

This is the first PR in a 3-step deployment process:

1. Add additional tables and migrate existing archived petitions
2. Change code relating to archived petitions to use the new models
3. Delete the old schema/code


app/models/archived/government_response.rb
```diff
@@ -0,0 +1,13 @@
+module Archived
+  class GovernmentResponse < ActiveRecord::Base
+    belongs_to :petition, touch: true
+
+    validates :petition, presence: true
+    validates :summary, length: { maximum: 200 }, allow_blank: true
+    validates :details, length: { maximum: 10000 }, allow_blank: true
```
pixeltrix: Yes

PR #597, opened 2017-06-15
Add proper tagging functionality

jamesmarkcook:
## Features added
- Tags on `Petition` and `ArchivedPetition`
- Radio button on the search form to specify what to search for
- Filter search results by tags (only petitions for now)
- Admin Site Settings (`Admin::Site`) page to manage allowed tags
- Only allow tags on petitions that are in the site settings
- Don't reload search parameters when filtering or selecting scope
- Show correct count on scope select box for filtered petitions
- Make tags case insensitive when searching. TAG 1 == tag 1. However display the original case for the user to select. Therefore if an admin sets an allowed tag to be HMRC, it will display as HMRC.

## Features/Niceties not added (yet)
- Filtering signature search results by tags. Thought this could be a separate ticket/PR.
- Adding JS tags autocompletion instead of displaying all tags at once
- Removing scopes with 0 results. Not sure if we want to do this?

## Technical details
- Introduce Taggable module to enable tag scopes on models that implement a tags column
- Introduce an AdminTagsValidation module to validate tags against allowed tags in admin site settings.
- Handle Signature searching on the Admin::SignaturesController
- Simplify Admin::PetitionsController to pass search params straight through to Browseable module. 
- Simplify Admin::SearchController to pass params straight to Admin::PetitionsController or Admin::SignaturesController
- Add TagFilter class to Browseable module in order to manage filtering of search results
- Search and filtering order in Browseable is now search -> filter results -> add scope


app/models/admin/site.rb
```diff
@@ -0,0 +1,50 @@
+require 'list_processor'
```
pixeltrix: Move `lib/list_processor.rb` to `app/lib/list_processor.rb` and drop this `require`

app/models/rate_limit.rb
```diff
@@ -1,6 +1,9 @@
 require 'ipaddr'
+require 'list_processor'
```
pixeltrix: Move `lib/list_processor.rb` to `app/lib/list_processor.rb` and drop this `require`

app/helpers/admin_helper.rb
```diff
@@ -59,6 +59,10 @@ def trending_ips?(since: 1.hour.ago, limit: 10)
     !trending_ips(since: since, limit: limit).empty?
   end
 
+  def admin_site_settings
+    @admin_site_settings ||= Admin::Site.first_or_create!
+  end
+
```
pixeltrix: Remove this and add a `before_action` in `Admin::AdminController` to load the settings object

app/models/admin/site.rb
```diff
@@ -0,0 +1,50 @@
...
+
+class Admin::Site < ActiveRecord::Base
```
pixeltrix: I think on reflection this would be better named as `Admin::Settings` - also use nested context do define the model, e.g.

``` ruby
module Admin
  class Settings < ActiveRecord::Base
    # implementation
  end
end
```

app/models/admin/site.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def duplicate_tags_not_allowed
+    tag_counts = petition_tags_map(for_comparison: true).each_with_object(Hash.new(0)) do |tag, counts|
+      counts[tag] += 1
+    end
+
+    duplicate_tags = tag_counts.collect { |tag, count| tag if count > 1 }.compact
+    if duplicate_tags.any?
+      error_message = "Duplicate tags not allowed: #{duplicate_tags.join(', ')}"
+      errors.add(:petition_tags, error_message)
```
pixeltrix: You should use a symbol here and the locale YAML file to build the error message. e.g.

``` ruby
errors.add(:petition_tags, :duplicate_tags, tags: duplicate_tags.join(', '))
```
``` yaml
en-GB:
  errors:
    messages:
      duplicate_tags: "Duplicate tags not allowed: %{tags}"
```

app/models/admin/site.rb
```diff
@@ -0,0 +1,50 @@
...
+  def no_petitions_have_deleted_tags
+    existing_tags = petition_tags_map(petition_tags_was, for_comparison: true)
+    new_tags = petition_tags_map(for_comparison: true)
+    deleted_tags = existing_tags - new_tags
+
+    petitions_with_deleted_tags = deleted_tags.each do |tag|
+      petitions = Petition.with_tag(tag)
+      if petitions.any?
+        error_message = "Tag '#{tag}' still being used on petitions: #{petitions.pluck(:id).join(', ')}"
+        errors.add(:petition_tags, error_message)
```
pixeltrix: Same here - use a symbol and the locale YAML file to build the error message. e.g.

``` ruby
errors.add(:petition_tags, :tag_in_use, tag: tag, petitions: petitions.pluck(:id).join(', '))
```
``` yaml
en-GB:
  errors:
    messages:
      tag_in_use: "Tag '%{tag}' is still being used on petitions: %{petitions}"
```

spec/models/petition_spec.rb
```diff
@@ -39,6 +39,16 @@
...
+  describe "included modules" do
+    it "includes AdminTagsValidation module" do
+      expect(Petition.included_modules).to include(AdminTagsValidation)
+    end
+
+    it "includes Taggable module" do
+      expect(Petition.included_modules).to include(Taggable)
+    end
+  end
+
```
pixeltrix: I don't think these tests are necessary

app/models/admin/site.rb
jamesmarkcook: 👍

app/models/concerns/taggable.rb
```diff
@@ -0,0 +1,43 @@
...
+    def acts_as_taggable_array_on(*tag_def)
+      self.tag_column = tag_def.first
+    end
+
+    def with_all_tags(tags)
+      where("array_lowercase(#{tag_column}) @> ARRAY[?]::varchar[]", downcase_tags(tags))
+    end
+
+    def with_tag(tag)
+      where("'#{tag.downcase}' = ANY (array_lowercase(#{tag_column}))")
```
NeilvB: While addressing @pixeltrix's comment on my PR (https://github.com/alphagov/e-petitions/pull/603), which extends this module, I've run `EXPLAIN` on this query and found that it used a sequential scan rather than the index.

I found https://stackoverflow.com/questions/22255503/why-isnt-my-postgresql-array-index-getting-used-rails-4/22265078 and found a way to change this to use the index: 

```
    def with_tag(tag)
      where("'{#{tag}}' <@ (array_lowercase(tags))")
    end
```

db/migrate/20170602154741_add_tags_to_petitions.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  def down
+    remove_column :petitions, :tags
+    remove_column :archived_petitions, :tags
+
+    execute <<-SQL
+      DROP INDEX IF EXISTS petitions_tag_lower
+      DROP INDEX IF EXISTS archived_petitions_tag_lower
+      DROP FUNCTION IF EXISTS array_lowercase
+    SQL
```
NeilvB: I think we need line endings and to pass in the argument type to the DROP FUNCTION here for this to execute

pixeltrix:
Closing in favour of #608 where the `intarray` extension is use to store tag IDs.

PR #604, opened 2017-06-28
Add jobs to archive current petitions

pixeltrix:
Also updates the archived petition pages to further match the the current petition pages (added debate outcomes and petition emails). Whilst it should be okay to deploy this to production, we can't run the archive tasks until we have the admin for archived petitions that allows us to send government responses, add debate outcomes, etc.


app/views/archived/petitions/show.html.erb
```diff
@@ -72,6 +72,73 @@
     </div>
 
     <div class="about-petitions">
+      <% if debate_outcome = @petition.debate_outcome %>
```
NeilvB: I haven't done a full like-for-like check but parts of this section appear similar to the parts of the view for non-archived petitions such as `_debate_threshold.html.erb` - is there enough commonality for us to reuse one or more of the partials here?

pixeltrix: Possibly, but we have to deal with both the old archived petitions and the new archived petitions and sometimes taking DRY to the limit isn't advisable.

app/models/petition.rb
```diff
@@ -343,7 +343,7 @@ def mark_petitions_as_debated!(date = Date.current)
     end
 
     def archived?
```
pixeltrix: Hmm, I guess to follow Rails naming convention it should be `archived_all?` similar to `update` vs. `update_all` and `destroy` vs. `destroy_all`.

I think maybe just having an `unarchived` scope and moving the check to the `Parliament` model would be clearer.

PR #619, opened 2017-08-18
Change black/white to neutral allow/block

pixeltrix:
Whilst the etymology of black/white listing isn't racially based, it does reinforce the stereotypes of white=good and black=bad so change them to the neutral terms allow/block.

H/T to https://twitter.com/secnerdette/status/898314208097427457 for the inspiration.


pixeltrix:
Normally renaming the columns like this would involve two PRs and deploys but we're currently on a parliamentary recess so no-one is signing petitions at the moment.

yahoopete:
:+1:

PR #636, opened 2017-12-01
Prevent boolean argument to prepared statements

pixeltrix:
When using hash conditions to `where`, Active Record will treat that
as an argument for the prepared statement even though it will never change in reality. This causes problems for our partial indexes on `creator` and `sponsor` columns in the signatures and the archived_signatures tables because of the following behaviour as documented in [PREPARE][1]:

> Prepared statements can use generic plans rather than re-planning with each set of supplied EXECUTE values. This occurs immediately for prepared statements with no parameters; otherwise it occurs only after five or more executions produce plans whose estimated cost average (including planning overhead) is more expensive than the generic plan cost estimate. Once a generic plan is chosen, it is used for the remaining lifetime of the prepared statement.

What this means in effect is that after each Active Record connection was used five times to fetch a petition creator it was defaulting to the generic query plan which in our case turned out to be a sequential scan of the signatures table.

By using an opaque Arel attribute node passed to `where` we can prevent Active Record from seeing that as an argument to a prepared statement so the plan used by PostgreSQL continues to use our partial index even after the statement has been used five times.

[1]: https://www.postgresql.org/docs/9.6/static/sql-prepare.html#SQL-PREPARE-NOTES

### DIFF
--- a/app/models/archived/petition.rb
+++ b/app/models/archived/petition.rb
@@ -16,7 +16,7 @@ module Archived
     belongs_to :parliament, inverse_of: :petitions, required: true
     belongs_to :locked_by, class_name: 'AdminUser'
 
-    has_one :creator, -> { where(creator: true) }, class_name: "Signature"
+    has_one :creator, -> { creator }, class_name: "Signature"
     has_one :debate_outcome, dependent: :destroy
     has_one :government_response, dependent: :destroy
     has_one :note, dependent: :destroy
@@ -24,7 +24,7 @@ module Archived
 
     has_many :emails, :dependent => :destroy
     has_many :signatures
-    has_many :sponsors, -> { where(sponsor: true) }, class_name: "Signature"
+    has_many :sponsors, -> { sponsors }, class_name: "Signature"
 
     validates :action, presence: true, length: { maximum: 150 }
     validates :background, length: { maximum: 300 }, allow_blank: true
--- a/app/models/archived/signature.rb
+++ b/app/models/archived/signature.rb
@@ -56,6 +56,14 @@ module Archived
       def validated
         where(state: VALIDATED_STATE)
       end
+
+      def creator
+        where(arel_table[:creator].eq(true))
+      end
+
+      def sponsors
+        where(arel_table[:sponsor].eq(true))
+      end
     end
 
     def get_email_sent_at_for(timestamp)
--- a/app/models/petition.rb
+++ b/app/models/petition.rb
@@ -65,7 +65,7 @@ class Petition < ActiveRecord::Base
   facet :tagged_in_moderation,         -> { tagged_in_moderation.by_most_recent_moderation_threshold_reached }
   facet :untagged_in_moderation,       -> { untagged_in_moderation.by_most_recent_moderation_threshold_reached }
 
-  has_one :creator, -> { where(creator: true) }, class_name: 'Signature'
+  has_one :creator, -> { creator }, class_name: 'Signature'
   accepts_nested_attributes_for :creator, update_only: true
 
   belongs_to :locked_by, class_name: 'AdminUser'
@@ -77,7 +77,7 @@ class Petition < ActiveRecord::Base
   has_one :rejection, dependent: :destroy
 
   has_many :signatures
-  has_many :sponsors, -> { where(sponsor: true) }, class_name: 'Signature'
+  has_many :sponsors, -> { sponsors }, class_name: 'Signature'
   has_many :country_petition_journals, dependent: :destroy
   has_many :constituency_petition_journals, dependent: :destroy
   has_many :emails, dependent: :destroy
--- a/app/models/signature.rb
+++ b/app/models/signature.rb
@@ -165,8 +165,12 @@ class Signature < ActiveRecord::Base
       scope.paginate(page: page, per_page: 50)
     end
 
+    def creator
+      where(arel_table[:creator].eq(true))
+    end
+
     def sponsors
-      where(sponsor: true)
+      where(arel_table[:sponsor].eq(true))
     end
 
     def subscribed


matthewd:
FWIW, I wouldn't trust that Arel node to stay opaque in future. Does plain `where(arel_table[:creator])` work?

(Though I'm also unclear on why that generic plan behaviour would cause what you're describing: the five queries should have *lower* specific-plan costs than the generic plan, no?)

pixeltrix:
> Does plain where(arel_table[:creator]) work?

Yes, it appears to.

> the five queries should have lower specific-plan costs than the generic plan, no?
Since the first row in the table is `creator = true` then the cost ranges from 0.0 for a sequential scan. However since the partial index query plan includes the planning overhead the cost range is non-zero.

It's trivial to reproduce this in `psql`: 

```
epetitions_production=> PREPARE query AS SELECT "signatures".* FROM "signatures" WHERE "signatures"."petition_id" = $1 AND "signatures"."creator" = $2 LIMIT 1;
PREPARE
epetitions_production=> EXPLAIN EXECUTE query(205476, true);
                                                           QUERY PLAN                                                           
--------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.28..8.30 rows=1 width=225)
   ->  Index Scan using index_signatures_on_petition_id_where_creator_is_true on signatures  (cost=0.28..8.30 rows=1 width=225)
         Index Cond: (petition_id = 205476)
(3 rows)

epetitions_production=> EXPLAIN EXECUTE query(205476, true);
                                                           QUERY PLAN                                                           
--------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.28..8.30 rows=1 width=225)
   ->  Index Scan using index_signatures_on_petition_id_where_creator_is_true on signatures  (cost=0.28..8.30 rows=1 width=225)
         Index Cond: (petition_id = 205476)
(3 rows)

epetitions_production=> EXPLAIN EXECUTE query(205476, true);
                                                           QUERY PLAN                                                           
--------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.28..8.30 rows=1 width=225)
   ->  Index Scan using index_signatures_on_petition_id_where_creator_is_true on signatures  (cost=0.28..8.30 rows=1 width=225)
         Index Cond: (petition_id = 205476)
(3 rows)

epetitions_production=> EXPLAIN EXECUTE query(205476, true);
                                                           QUERY PLAN                                                           
--------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.28..8.30 rows=1 width=225)
   ->  Index Scan using index_signatures_on_petition_id_where_creator_is_true on signatures  (cost=0.28..8.30 rows=1 width=225)
         Index Cond: (petition_id = 205476)
(3 rows)

epetitions_production=> EXPLAIN EXECUTE query(205476, true);
                                                           QUERY PLAN                                                           
--------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.28..8.30 rows=1 width=225)
   ->  Index Scan using index_signatures_on_petition_id_where_creator_is_true on signatures  (cost=0.28..8.30 rows=1 width=225)
         Index Cond: (petition_id = 205476)
(3 rows)

epetitions_production=> EXPLAIN EXECUTE query(205476, true);
                                                   QUERY PLAN                                                   
----------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.43..4.96 rows=1 width=225)
   ->  Index Scan using index_signatures_on_petition_id on signatures  (cost=0.43..5659.04 rows=1249 width=225)
         Index Cond: (petition_id = $1)
         Filter: (creator = $2)
(4 rows)
```

pixeltrix:
Hmm, looking at those scores, it's the top end of the range which is higher on the partial index plan so I have know idea what Postgres is doing? 🤷‍♂️

matthewd:
Ouch! Does it behave differently given a much lower id?

Actually.. does the new fixed-boolean form also kick over to a generic plan after 5 runs with that same id? (Still using the good index, of course, but switching from `Index Cond: (petition_id = 205476)` to `Index Cond: (petition_id = $1)`) If so, what do those costs look like?

pixeltrix:
> Ouch! Does it behave differently given a much lower id?

The petition ids are starting from 200000 so that is a relatively low value 😉 

> Actually.. does the new fixed-boolean form also kick over to a generic plan after 5 runs with that same id?

Yes, it does fallback to the generic plan but the costs are identical:

```
epetitions_production=> PREPARE query AS SELECT "signatures".* FROM "signatures" WHERE "signatures"."petition_id" = $1 AND "signatures"."creator" LIMIT 1;
PREPARE

epetitions_production=> EXPLAIN EXECUTE query(205476);
                                                           QUERY PLAN                                                           
--------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.28..8.30 rows=1 width=225)
   ->  Index Scan using index_signatures_on_petition_id_where_creator_is_true on signatures  (cost=0.28..8.30 rows=1 width=225)
         Index Cond: (petition_id = 205476)
(3 rows)

epetitions_production=> EXPLAIN EXECUTE query(205476);

...

epetitions_production=> EXPLAIN EXECUTE query(205476);
                                                           QUERY PLAN                                                           
--------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.28..8.30 rows=1 width=225)
   ->  Index Scan using index_signatures_on_petition_id_where_creator_is_true on signatures  (cost=0.28..8.30 rows=1 width=225)
         Index Cond: (petition_id = $1)
(3 rows)
```

I guess it's saying that the specific plans are no better.

matthewd:
So definitely not anything to do with the particular id value being slightly better/worse than an average / parameterized value.

I *think* that means it's expecting `creator` to be a lot less selective than it obviously is... though even if it thought there was a 90% chance the first row would pass the filter, it still seems like the partial index would be the safer bet. :confused:

I think I'd drop by `#postgresql` for a second opinion -- avoiding the possibility of an overly-generic plan is a pretty safe bet, but I wonder whether there are some stats or something to be prodded.

Even if it thought the `creator` condition was going to select *every* row, I'd expect the total retrieval cost from the specialized index to be ~ the same, instead of double.

---

Interestingly, I think this problem would go away on Rails 5: the limit is parameterized, so the generic plan would be much closer to the full 5659.04 cost.

pixeltrix:
> Interestingly, I think this problem would go away on Rails 5: the limit is parameterized, so the generic plan would be much closer to the full 5659.04 cost.

Yes, just checked - it keeps on using the specific plan with a parameterised limit value.

pixeltrix:
@matthewd just using `arel_table[:creator]` doesn't work because it doesn't set `creator` to `true` when creating the signature through the `has_one` association. 😞

matthewd:
Ah, right. At least https://github.com/alphagov/e-petitions/pull/636#issuecomment-348515873 suggests that any future change will be okay: if the arel node gets parameterized, so will the limit. 👍🏻

PR #652, opened 2018-05-14
apply theme colour to mobile Chrome user interface

willp-bl:
Note: I've not actually tested this but it is a small enough change

### DIFF
--- a/app/views/layouts/admin.html.erb
+++ b/app/views/layouts/admin.html.erb
@@ -9,6 +9,7 @@
   <%= yield :js %>
   <meta name="viewport" content="width = device-width" />
   <meta name="robots" content="noindex" />
+  <meta name="theme-color" content="#080">
 </head>
 <body class="admin">
   <%= render '/admin/shared/header' %>
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -19,6 +19,7 @@
     <% if noindex_page? %>
       <meta name="robots" content="noindex" />
     <% end %>
+      <meta name="theme-color" content="#080">
   </head>
   <body>
     <% if show_cookie_message? %>


yahoopete:
@pixeltrix this will be hard to view locally, could we merge and check on preview please?

app/views/layouts/admin.html.erb
```diff
@@ -9,6 +9,8 @@
   <%= yield :js %>
   <meta name="viewport" content="width = device-width" />
   <meta name="robots" content="noindex" />
+  <%# theme-color is from app/assets/stylesheets/petitions/_colours.scss:$house-of-commons-green %>
+  <meta name="theme-color" content="#080">
```
pixeltrix: @willp-bl this only applies to home page style web apps right? No-one is going to be adding the admin site to their phone's home page since it doesn't really work on mobile so you might as well remove it.

app/views/layouts/application.html.erb
```diff
@@ -19,6 +19,8 @@
     <% if noindex_page? %>
       <meta name="robots" content="noindex" />
     <% end %>
+      <%# theme-color is from app/assets/stylesheets/petitions/_colours.scss:$house-of-commons-green %>
```
pixeltrix: I don't think the comment is really necessary - can you remove it please.

app/views/layouts/admin.html.erb
willp-bl: it will be visible in the normal mobile chrome browser when you look at any admin pages there, same as the screenshots above, let me know if you still want it removed

app/views/layouts/application.html.erb
willp-bl: sure

willp-bl:
I've removed the code comments as requested, thanks for reviewing 👍

pixeltrix:
@yahoopete I've put this on https://preview.epetitions.website - if you can find someone with an Android phone and add it to their home screen you can see what it looks like.

yahoopete:
Yep - I have said Android phone... looks great.

PR #657, opened 2018-06-01
add dev startup script that uses a dockerised postgres

willp-bl:
when this is run the `db/structure.sql` file is modified and I am not
sure why


pixeltrix:
Rather than just a dockerised database it would better be a full setup using docker compose as I did in #420 but we need to revisit the current infrastructure setup soon so I think we'll hold off on this until that's settled.

willp-bl:
👍

PR #686, opened 2019-03-22
Increase count auto-refresh period to 10 minutes

Jonty:
Given petitions are being updated less frequently, there's no need to poll for a count update every ten seconds. This should substantially reduce load generated by people leaving the petition page open.


pixeltrix:
Thanks for the PR @Jonty, we will be adjusting this value once we've finished the refactoring of signature incrementing but I think deploying now wouldn't actually help since people would just mash the refresh button instead and make a much heavier request.

PR #706, opened 2019-04-01
Feature/add simplecov

etagwerker:
Hey @pixeltrix, 

This is not solving an issue but more of a _nice to have_. 

Screen Shot 2019-03-31 at 9 42 45 PM

I added `SimpleCov` as a development dependency. It's useful to find what parts of the app are least covered (ie. libraries)

Please check it out and let me know if you think it's useful.

Best,
Ernesto

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 # Ignore logfiles and tempfiles.
 /log/*
 /tmp/*
+/coverage/*
 
 # Ignore precompiled assets
 /public/assets
--- a/Gemfile
+++ b/Gemfile
@@ -38,6 +38,7 @@ gem 'redcarpet'
 gem 'aws-sdk', '~> 2.0'
 
 group :development, :test do
+  gem 'simplecov'
   gem 'brakeman', require: false
   gem 'bundler-audit', require: false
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -34,3 +34,6 @@
 
 # Add back any keep files that might have been excluded by the above
 !.keep
+
+# SimpleCov
+coverage
```
pixeltrix: Small nitpick - can you change this to:

```
# Ignore coverage reports from SimpleCov
/coverage/*
```

and put it on line 9 before `# Ignore logfiles and tempfiles.` - the `!.keep` should be the last thing and file is sort of alphabetical

pixeltrix:
Happy to merge as long as it's a opt-in and not on by default (which it appears to be 👍 )

.gitignore
etagwerker: @pixeltrix Sure thing. No problem.

etagwerker:
@pixeltrix Cool, just updated the changes based on your feedback re: `.gitignore`

pixeltrix:
@etagwerker thanks! 👍

PR #730, opened 2019-05-20
Update the colour for MHCLG

AlanGabbianelli:
The colour is wrong in a few places and they've asked us to update it.

Trello card: 
https://trello.com/c/wqKT3TpE/952-change-mhclgs-brand-colour


pixeltrix:
@AlanGabbianelli we don't use the department brand colours anywhere within the Petitions service - either on the frontend or backend but happy to merge

PR #740, opened 2020-01-08
Gracefully handle when signatures have never been counted

pixeltrix:
When a new developer boots the app for the first time the counts have never been updated so the job fails. Fix this by defaulting to the earliest validated signature and if there are no validated signatures then just record the current time window and exit.

### DIFF
--- a/spec/jobs/update_signature_counts_job_spec.rb
+++ b/spec/jobs/update_signature_counts_job_spec.rb
@@ -1,12 +1,14 @@
 require 'rails_helper'
 
 RSpec.describe UpdateSignatureCountsJob, type: :job do
-  let(:current_time) { Time.current.change(usec: 0) }
   let(:interval) { 30 }
+  let(:current_time) { Time.current.change(usec: 0) }
   let(:scheduled_time) { interval.seconds.since(current_time) }
+  let(:previous_count_updated_at) { current_time - (2 * interval).seconds }
+  let(:current_count_updated_at) { current_time - interval.seconds }
 
   before do
-    Site.signature_count_updated_at!(current_time - 60.seconds)
+    Site.signature_count_updated_at!(previous_count_updated_at)
     allow(Site).to receive(:signature_count_interval).and_return(interval)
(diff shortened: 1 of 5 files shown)


spec/jobs/update_signature_counts_job_spec.rb
```diff
@@ -331,6 +331,68 @@
...
+          Site.signature_count_updated_at!(nil)
+        end
+
+        context "and there are no validated signatures" do
+          it "updates Site#signature_count_updated_at" do
+            expect {
+              described_class.perform_now(current_time.iso8601)
+            }.to change {
+              Site.signature_count_updated_at
+            }.from(nil).to(current_time - 30.seconds)
```
ayghor: wait `30.seconds`?

ayghor: is this the update interval? maybe it's better to say `interval.seconds`?

pixeltrix: The signature counting window runs one window behind to ensure that all pending transactions have been committed since the validated_at timestamp is generated in Ruby not the database. If we didn’t then a pending signature validation could be missed.

The reason for not doing `interval.seconds` is it’s tie the expected value to the setup and if for some reason it was changed to something else accidentally the test wouldn’t fail. I alway prefer literal values for expectations wherever possible.

pixeltrix: This is the historical counting PR: https://github.com/alphagov/e-petitions/pull/712

ayghor: > The reason for not doing interval.seconds is it’s tie the expected value to the setup and if for some reason it was changed to something else accidentally the test wouldn’t fail. I alway prefer literal values for expectations wherever possible.

but even though it was defined on the spec itself?
```ruby
RSpec.describe UpdateSignatureCountsJob, type: :job do
  let(:current_time) { Time.current.change(usec: 0) }
  let(:interval) { 30 }
```
indeed it could be changed accidentally, but so could other `let()` definitions, no?

PR #745, opened 2020-02-13
Refactor some petition specs

ayghor:


ayghor:
:sparkles: :sparkles: :sparkles:

PR #763, opened 2020-04-24
Similar petitions grammar fix

joelanman:
Before

> But, there might be a similar petition has used different words.

After

> But there might be a similar petition that has used different words.


pixeltrix:
Some weird selenium/chromedriver issue with details/summary elements - unrelated to the change here

pixeltrix:
Deployed - thanks for the fix 👍

PR #767, opened 2020-08-13
Apply ip/domain allow/block lists to petition creation

pixeltrix:
The thank you page is moved to the collection routing level since it's not guaranteed that we have an id anymore but we don't need it anyway.

### DIFF
--- a/spec/controllers/petitions_controller_spec.rb
+++ b/spec/controllers/petitions_controller_spec.rb
@@ -69,7 +69,7 @@ RSpec.describe PetitionsController, type: :controller do
         end
 
         expect(petition.creator).not_to be_nil
-        expect(response).to redirect_to("https://petition.parliament.uk/petitions/#{petition.id}/thank-you")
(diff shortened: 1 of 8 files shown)


spec/controllers/petitions_controller_spec.rb
```diff
@@ -272,6 +272,24 @@
...
+      before do
+        rate_limit.update!(blocked_ips: "0.0.0.0")
+      end
+
+      it "should not create a new petition and a signature" do
+        perform_enqueued_jobs do
+          post :create, params: { stage: "replay_email", petition_creator: params }
+        end
+
+        expect(petition).to be_nil
```
pudinha, line 289: I'd put a `expect {}.not_to change { Petition.count }` here, the `"Save the planet"` could fail to match for other reasons. I  was also expecting a check on signature since it's on the description

pudinha: I mean `find_by_action("Save the planet")` could fail to match

PR #770, opened 2020-08-13
Rate limit creator signatures across petitions

pixeltrix:
TODO: add a separate rate limit for petition creation.

### DIFF
--- a/features/charlie_creates_a_petition.feature
+++ b/features/charlie_creates_a_petition.feature
@@ -201,3 +201,20 @@ Scenario: Charlie creates a petition when blocked
   When I press "Yes – this is my email address"
   Then a petition should not exist with action: "The wombats of wimbledon rock.", state: "pending"
   And a signature should not exist with email: "womboid@wimbledon.com", state: "pending"
+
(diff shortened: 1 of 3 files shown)


features/charlie_creates_a_petition.feature
```diff
@@ -201,3 +201,20 @@ Scenario: Charlie creates a petition when blocked
   When I press "Yes – this is my email address"
   Then a petition should not exist with action: "The wombats of wimbledon rock.", state: "pending"
   And a signature should not exist with email: "womboid@wimbledon.com", state: "pending"
+
+Scenario: Charlie creates a petition when his IP address is rate limited
+  Given the burst rate limit is 1 per minute
+  And there are no allowed IPs
+  And there are no blocked IPs
+  And there are 2 petitions created from this IP address
```
pudinha, line 209: two petitions, not one?

pixeltrix: Yes, because the creator signature hasn't been created yet so will pass the `<` check if it's only one.

PR #787, opened 2021-04-25
Improve searching of petitions

pixeltrix:
By using the btree_gin extension and merging the indexes on the various columns into a single index combined with state, debate_state and parliament_id for archived petitions we can achieve a significant performance gain of more than 10x when searching arbitrary text.

This also adds in the petition id to the index so people can now search by that on the front end - it will even return cases where people refer to the petition by id in other petition text.


RokeJulianLockhart:
@pixeltrix, good luck! LLM-generated anything is always a little hairy-scary.

PR #810, opened 2021-11-24
Remove mention of GA cookies from privacy notice

cdccollins:
- GA is being removed from the site, so no longer need to mention them
in the privacy notice
- Also update when the privacy notice was last updated


app/views/pages/privacy.html.erb
```diff
@@ -6,7 +6,7 @@
 
   <h2 id="privacy-heading">Privacy notice: how we use your data</h2>
 
-  <p>Last update: March 17 2021</p>
+  <p>Last update: November 2021</p>
```
pixeltrix, line 9: Should include the day

pixeltrix (approved):

PR #918, opened 2024-03-15
Upgrade to the latest version of Rails

pixeltrix:
Also upgrades the cucumber environment which was long overdue.


config/application.rb
```diff
@@ -23,29 +22,35 @@
 module Epets
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 6.1
+    config.load_defaults 7.0
```
benjamineskola, line 25: Should this be 7.1?

PR #943, opened 2024-06-06
Upgrade to Debian Stable and Ruby 3.2

benjamineskola:


pixeltrix:
Probably want to upgrade the GHA config to use Ruby 3.2 and ubuntu-24.04

pixeltrix:
All the errors seem related to the Active Storage image on the debate outcome models

benjamineskola:
> All the errors seem related to the Active Storage image on the debate outcome models

I can't reproduce this locally (in docker) — wondering if there's some difference between Debian and Ubuntu?

.github/actions/setup/action.yml
```diff
@@ -8,6 +8,11 @@ runs:
       run: |
         sudo apt-get -yqq install libpq-dev
 
+    - name: Install image processing tools
+      shell: bash
+      run: |
+        sudo apt-get -yqq install imagemagick libvips
```
pixeltrix, line 14: Interesting this is not in the default setup - seems to be a change in the 24.04 runner vs the 22.04 runner

pixeltrix (approved):
We probably want to upgrade bundler as well, but that'll need changing in the server config as well

PR #944, opened 2024-06-11
Migrate primary keys to bigint

benjamineskola:


benjamineskola:
This has now been fully applied (manually) in a snapshot on preview.

PR #950, opened 2024-06-18
Feature/update parliament endpoints

Rhian:
This PR adds two new endpoints: a Parliaments index, and an individual Parliament response, defined by period, which contains a list of constituencies for the parliament, along with other attributes.


db/migrate/20240615110144_create_and_populate_join_table_constituencies_parliaments.rb
```diff
@@ -0,0 +1,8 @@
+class CreateAndPopulateJoinTableConstituenciesParliaments < ActiveRecord::Migration[7.1]
+  def change
+    create_join_table :constituencies, :parliaments do |t|
+      t.string :constituency_external_id
```
pixeltrix, line 4: This adds a third column - the join table should just have two columns. If you can't get the right settings by changing the options, just use the regular `create_table` with an `id: false` option.

app/models/constituency.rb
```diff
@@ -105,4 +106,20 @@ def mp_url
...
+    end_date.nil? && start_date < parliament.dissolution_at
+  end
+
+  def overlaps_previous_parliament_when_archived?(parliament)
+    end_date.present? && parliament.dissolution_at.present? && start_date < parliament.dissolution_at && end_date > parliament.opening_at
+  end
+
+  def intersects_parliament?(parliament)
+    overlaps_current_parliament?(parliament) || overlaps_previous_parliament_when_current?(parliament) || overlaps_previous_parliament_when_archived?(parliament)
+  end
```
pixeltrix, line 124: I'm still looking at this - I'm sure it can be done with less steps

pixeltrix (changes requested):

db/migrate/20240615110144_create_and_populate_join_table_constituencies_parliaments.rb
Rhian: Now we are running regular updates on the import, this can be dropped. I'd originally included it in case constituency ids changed and we had to index on `external_id`

app/models/constituency.rb
```diff
@@ -8,6 +8,7 @@ class Constituency < ActiveRecord::Base
   belongs_to :region, primary_key: :external_id, optional: true
   has_many :signatures, primary_key: :external_id
   has_many :petitions, through: :signatures
+  has_and_belongs_to_many :parliaments
```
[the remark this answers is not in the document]
Rhian: I've removed `external_id` from the join table, so presumably it's OK to not index as it should automatically index on `constituency.id`?

db/migrate/20240613095954_add_period_to_parliament.rb
```diff
@@ -0,0 +1,15 @@
+class AddPeriodToParliament < ActiveRecord::Migration[7.1]
+  def change
+    add_column :parliaments, :period, :string
```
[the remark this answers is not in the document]
Rhian: Amended with one small update to set `Time.now` as a variable and interpolate it in the query.

app/jobs/fetch_constituencies_job.rb
```diff
@@ -29,6 +29,8 @@ def perform
           end
 
           constituency.save!
+
+          update_constituencies_parliaments(constituency)
```
pixeltrix, line 33: I think we need to do this outside the constituencies loop and do it for each parliament, e.g.

``` ruby
constituencies.each do |external_id, name, ons_code, start_date, end_date|
# existing code
end

parliaments.each do |parliament|
  parliament.constituency_ids = Constituency.for_parliament(parliament).external_ids
end

class Constituency < ApplicationRecord
  class << self
    def for_parliament(parliament)
      if parliament.archived? || parliament.dissolved?
        between(parliament.opening_at, parliament.dissolution_at)
      else
        current
      end
    end

    def between(opening_at, dissolution_at)
      where(arel_table[:start_date].lteq(opening_at).and(arel_table[:end_date].gteq(dissolution_at).or(arel_table[:end_date].eq(nil))))
    end

    def external_ids
      pluck(:external_id)
    end
  end
end
```

We'll need to add some tests for these scopes - check the following:

1. Parliaments that have been archived
2. Parliaments that have been dissolved
3. Parliaments that are current before opening
4. Parliaments that are current after opening

For each of those scenarios, check that

1. Constituencies ended before parliament opened are excluded
2. Constituencies started after parliament opened are excluded
3. Constituencies started before parliament opened and ended after dissolution are included
4. Constituencies started before parliament opened and not ended are included

app/jobs/fetch_constituencies_job.rb
```diff
@@ -62,4 +64,18 @@ def regions
...
   end
+
+  def parliaments
+    Parliament.all
+  end
+
+  def update_constituencies_parliaments(constituency)
+    parliaments.each do |parliament|
+      if constituency.intersects_parliament?(parliament)
+        ConstituenciesParliament.find_or_create_by!(constituency_id: constituency.id, parliament_id: parliament.id)
```
pixeltrix, line 75: This is why you're needing to specify the model - you're using as a model and the default class that Rails creates is something like `Parliament::Constituencies_HABTM`. If you do what I suggest in the other comment you'll be able to delete the model as all the work is done through the association.

pixeltrix (changes requested):

app/controllers/parliaments_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+      format.json
+    end
+  end
+
+  def show
+    @parliament = Parliament.archived.find_by!(period: params[:period])
+
+  respond_to do |format|
+     format.json
+    end
```
pixeltrix, line 17: Indenting of `respond_to` is off here

spec/support/parliament.rb
```diff
@@ -5,7 +5,9 @@
   end
 
   config.before(:each) do |example|
-    Parliament.reset!(government: "TBC", opening_at: 2.weeks.ago)
-  end
+      unless example.metadata[:skip_before_hook]
```
pixeltrix, line 8: Indenting is off here as well

pixeltrix (approved):
Just some formatting issues 👍🏻

benjamineskola (approved):

PR #954, opened 2024-06-27
Add parliament details to archived_petition json

Rhian:
This PR adds the details of the parliament in which the petition was raised to the `archived/petitions/:id.json` view only.


app/views/archived/petitions/_petition.json.jbuilder
```diff
@@ -5,6 +5,13 @@ json.links do
   json.self archived_petition_url(petition, format: :json)
 end if defined?(is_collection)
 
+json.parliament do
+  json.period petition.parliament.period
```
pixeltrix, line 9: We should include the `dissolution_at` timestamp which is in the parliaments/show template

pixeltrix (approved):

PR #957, opened 2024-07-11
Don't run tasks when parliament has dissolved

pixeltrix:
When parliament has dissolved there is no need to run maintenance tasks.


lib/tasks/epets.rake
```diff
@@ -74,6 +74,7 @@ namespace :epets do
     desc "Start the signature count updater if it's not running"
     task :signature_counts => :environment do
       Task.run("epets:site:signature_counts", 10.minutes) do
+        break if Parliament.closed?(48.hours.ago)
```
pixeltrix, line 77: We still run the tasks for 48 hours after dissolution as there is a 24 hour window for signatures to be verified.

PR #1144, opened 2025-02-10
Demonstration of using pgvector to improve searching

pixeltrix:
By using embeddings generated from a LLM we can hope to gain better results when search for existing petitions and also when people are search generally.


pixeltrix:
Improved version has been merged to a development branch for release at a future date

PR #1176, opened 2025-03-13
Fix flaky specs with WebMock disappearing stubs

pixeltrix:
There appears to be a race condition or timing issue where WebMock stubs are disappearing causing tests to fail. To temporarily work around the issue pin the version of chrome/chromium.


pixeltrix:
Wasn't the problem - need to expect the content of the response to ensure the completion of the page before moving on.

PR #1185, opened 2025-03-22
Establish a baseline using Chrome 128

pixeltrix:


pixeltrix:
Closing as it seems browser-actions/setup-chrome is broken

PR #1254, opened 2025-07-14
Alternative approach to better searching using Informers gem

pixeltrix:


pixeltrix:
Closing because we'd need EC2 instance types that can run ONXX models

PR #1276, opened 2025-08-03
Fix warnings generated by the new herb linter

pixeltrix:
Experiment in trying out the new Herb tools: https://herb-tools.dev/


pixeltrix:
The bulk of the changes stem from Herb not liking multiple HTML ids even when they're on different branches: https://github.com/marcoroth/herb/issues/273

PR #1413, opened 2026-01-08
Remove X share link.

dracos:
The UK technology secretary Liz Kendall said on Tuesday: "We cannot and will not allow the proliferation of these demeaning and degrading images, which are disproportionately aimed at women and girls. Make no mistake, the UK will not tolerate the endless proliferation of disgusting and abusive material online. We must all come together to stamp it out."
X has created and is running a piece of computer software that generates, hosts, and publishes these images. The government does not need to prioritise the X service as one of the sharing options at the bottom of every page where these share links are shown; the landscape for sharing information is not the same as it was when this was created.
(Might be easier to drop the whole sharing bar, given browsers generally have good sharing built in now and people know how to share things.)


pixeltrix:
Closing this as the new website redesign uses the [Web Sharing API][1] which was implemented in 6586693

[1]: https://developer.mozilla.org/en-US/docs/Web/API/Web_Share_API

PR #1668, opened 2026-08-19
Show petition title, creator, and text in signing flows

Jonty:
Currently during the sponsor and signing flows there's no information about the petition you are actually signing, which means people can't be sure what they are about to sign if someone sends them a link directly!

This introduces a header to all sponsor and signing flows that lists the petition title and creator. If the petition is already public it also links to the petition page, and if it is still in the sponsorship phase it shows the petition text under an expander instead.

My ruby is extremely rusty so this may need someone with actual talent to finesse it, and while I've tried to copy the standard design patterns this may not be up to scratch. I did my best!

### This is what it looks like in the sponsor flow (after being expanded):
image

### And this is what it looks like in the regular signature flow:
image

### DIFF
--- a/app/views/signatures/_petition_context.html.erb
+++ b/app/views/signatures/_petition_context.html.erb
@@ -1,0 +1,28 @@
+<p class="margin-bottom-2">
+  You’re signing the petition:
+  <% if full_text %>
+    <strong><%= petition.action %></strong>
+  <% else %>
+    <strong><%= link_to "#{petition.action}", petition_path(petition), target: "_blank" %></strong>
+  <% end %>
+</p>
+
+<% if (creator = petition.creator) && !creator.anonymized? %>
+  <span class="form-hint margin-bottom-7">
+    Created by <strong><%= creator.name %></strong>
+  </span>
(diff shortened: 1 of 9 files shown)


pixeltrix:
Thanks for the PR - a couple of points:

1. We don't show any UGC during the sponsor flow because that has not gone via moderation and the sponsor url could be (and often is) posted on social media despite the original intention being that the creator would share the petition via email which does contain the petition details. That way if there were any kind of negative or abusive content would only come from the petition creator and not be visible on the website. I'll bring it up at the next standup but I suspect that we'd want to keep the sponsor flow free of UGC.

2. People linking to the new signature page instead of the petition page is something we've noticed before and sometimes it can be nefarious in that people will link an image of one petition to an opposing petition's new signature page. I wonder if an addition of redirecting to the petition page if the referrer header isn't set to the correct petition might be an approach we can take here as well as adding summary details about the petition? I appreciate the referrer header can be manipulated but it would help prevent people linking to the new signature page directly and using misleading link text/images. What do you think?

